PR #21, opened 2016-12-29
Feature/8

Mobe91:


ginccc:
still relevant, but needs to be integrated into the latest, closed for now but branch will be kept in place for further reference!

PR #78, opened 2017-09-01
Facebook Quick Replies + Callbackmodule fixed

rolandpickl:
Because of the json serialization it was necessary to create a Holder Object in the Callbackmodule. This works now.
Also in this branch is the implementation of the quick replies for Facebook. Currently only text quickreplies are possible.

---
This change is [Reviewable](https://reviewable.io/reviews/labsai/eddi/78)

### DIFF
--- a/conversationcallback-definition/src/main/java/ai/labs/callback/model/ConversationDataResponseHolder.java
+++ b/conversationcallback-definition/src/main/java/ai/labs/callback/model/ConversationDataResponseHolder.java
@@ -1,0 +1,16 @@
+package ai.labs.callback.model;
+
+import ai.labs.memory.model.ConversationMemorySnapshot;
+import lombok.Getter;
+import lombok.Setter;
+
+/**
+ * Created by rpi on 31.08.2017.
+ */
+@Getter
+@Setter
+
+public class ConversationDataResponseHolder {
+    private ConversationMemorySnapshot conversationMemorySnapshot;
+}
+
--- a/conversationcallback-impl/src/main/java/ai/labs/callback/http/ConversationCallback.java
+++ b/conversationcallback-impl/src/main/java/ai/labs/callback/http/ConversationCallback.java
@@ -48,9 +48,11 @@ public class ConversationCallback implements IConversationCallback {
 
             dataResponse.setHttpCode(httpResponse.getHttpCode());
             dataResponse.setHeader(httpResponse.getHttpHeader());
+            log.info("content as string: " + httpResponse.getContentAsString());
             dataResponse.setConversationMemorySnapshot(
                     jsonSerialization.deserialize(httpResponse.getContentAsString(),
                             ConversationMemorySnapshot.class));
+
 
         } catch (IRequest.HttpRequestException | IOException e) {
             log.error(e.getLocalizedMessage(), e);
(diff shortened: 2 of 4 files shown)


ginccc:
*[conversationcallback-definition/src/main/java/ai/labs/callback/model/ConversationDataResponseHolder.java, line 13 at r1](https://reviewable.io:443/reviews/labsai/eddi/78#-KtC8mtjnt0XPZbBiL9D:-KtC8mtjnt0XPZbBiL9E:b6nn2ub) ([raw file](https://github.com/labsai/eddi/blob/eccc072b5087f2c92eaa1947dcde3f6a167866e4/conversationcallback-definition/src/main/java/ai/labs/callback/model/ConversationDataResponseHolder.java#L13)):*
> ```Java
> @Setter
> 
> public class ConversationDataResponseHolder {
> ```

Why was this necessary? 

Returning the ConversationMemorySnapshot as in ai/labs/rest/rest/IRestBotEngine.java:39 also works... what was the problem here?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/labsai/eddi/78)*

ginccc:
*[facebookmessenger-impl/src/main/java/ai/labs/facebookmessenger/endpoint/FacebookEndpoint.java, line 191 at r1](https://reviewable.io:443/reviews/labsai/eddi/78#-KtC9D2ZNF8epL7MSa_9:-KtC9D2ZNF8epL7MSa_A:byoye04) ([raw file](https://github.com/labsai/eddi/blob/eccc072b5087f2c92eaa1947dcde3f6a167866e4/facebookmessenger-impl/src/main/java/ai/labs/facebookmessenger/endpoint/FacebookEndpoint.java#L191)):*
> ```Java
>             log.debug("response: {}", httpResponse.getContentAsString());
>             final List<String> output = getOutputText(httpResponse.getContentAsString());
>             final List<Map<String, String>> quickReplies = getQuickReplies(httpResponse.getContentAsString());
> ```

This won't work well if there are more outputs with quickReplies

each output:someKey could have a quickReply:someKey to it that should be send to the client in order..

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/labsai/eddi/78)*

rolandpickl:
The problem is that Facebook only allows one line of quick replies. So even
if you have more chat bubbles, you can only have one quick reply line. This
is represented here.
I can do ordering by key, but I don't think that this will be easy to use
or understand for that matter.

On 4 Sep 2017 3:52 PM, "Gregor Jarisch" <notifications@github.com> wrote:

*facebookmessenger-impl/src/main/java/ai/labs/facebookmessenger/endpoint/FacebookEndpoint.java,
line 191 at r1
<https://reviewable.io:443/reviews/labsai/eddi/78#-KtC9D2ZNF8epL7MSa_9:-KtC9D2ZNF8epL7MSa_A:byoye04>
(raw file
<https://github.com/labsai/eddi/blob/eccc072b5087f2c92eaa1947dcde3f6a167866e4/facebookmessenger-impl/src/main/java/ai/labs/facebookmessenger/endpoint/FacebookEndpoint.java#L191>):*

            log.debug("response: {}", httpResponse.getContentAsString());
            final List<String> output =
getOutputText(httpResponse.getContentAsString());
            final List<Map<String, String>> quickReplies =
getQuickReplies(httpResponse.getContentAsString());

This won't work well if there are more outputs with quickReplies

each output:someKey could have a quickReply:someKey to it that should be
send to the client in order..
------------------------------

*Comments from Reviewable
<https://reviewable.io:443/reviews/labsai/eddi/78>*

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/labsai/EDDI/pull/78#issuecomment-326968643>, or mute
the thread

PR #104, opened 2018-04-30
Feature/output delay

ginccc:
In order to be backwards compatible, introduced a new output type 'delay' which shall delay the output for X milliseconds.

```
{
          "type": "delay",
          "valueAlternatives": [
            "3000"
          ]
        },
        {
          "type": "text",
          "valueAlternatives": [
            "some text here"
          ]
        }
```

---
This change is [Reviewable](https://reviewable.io/reviews/labsai/eddi/104)


rolandpickl:
:lgtm: but there is one thing: the bots now have a maximum number of concurrent requests with the definition of the system parameter. This could lead to blocked communications. We should try to find a solution that auto scales ...

---

Reviewed 4 of 5 files at r1, 1 of 1 files at r2.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io/reviews/labsai/eddi/104#-:-LBN5yQW5iFZ7gYFMSOe:bfkx4sz)*

PR #136, opened 2019-05-29
Facebook channel bugfix

rolandpickl:
just a tiny change in the facebook channel. The value that is written in the quick reply on facebook is now also the value that is sent back to eddi. This was done differently, but made more troubles than it was worth. This change does not warrant a new version

---
This change is [Reviewable](https://reviewable.io/reviews/labsai/eddi/136)


ginccc:
:lgtm:

rolandpickl:
facebook changes merged

PR #152, opened 2019-11-20
Add ability to configure JVM additional options. Add some config for production

kkorsakov:
Options are written free-style so take care of syntax.

Configure some production-grade options for better diagnosis (GC logs, taking heap dump on OOM, etc). Configure jdk.nio.MaxCachedBufferSize to not to overrun non-heap memory on large number of concurrent http threads.

---
This change is [Reviewable](https://reviewable.io/reviews/labsai/eddi/152)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,11 +1,10 @@
 version: '3'
 services:
   eddi:
-    image: labsai/eddi:4.8.0-kkorsakov
+    image: labsai/eddi:latest
     ports:
       - "7070:7070"
       - "7443:7443"
-      - "9001:9001"
     links:
       - mongodb
     depends_on:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,20 +1,23 @@
 version: '3'
 services:
   eddi:
-    image: labsai/eddi:latest
+    image: labsai/eddi:4.8.0-kkorsakov
     ports:
       - "7070:7070"
       - "7443:7443"
+      - "9001:9001"
     links:
       - mongodb
     depends_on:
       - mongodb
     environment:
       EDDI_VERSION: 4.8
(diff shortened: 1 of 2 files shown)


ginccc:
:lgtm:

docker-compose.yml
```diff
@@ -1,19 +1,24 @@
 version: '3'
 services:
   eddi:
-    image: labsai/eddi:latest
+    image: labsai/eddi:4.8.0-kkorsakov
```
ginccc: this needs to be reverted back to 'latest'
```suggestion
    image: labsai/eddi:latest
```

docker-compose.yml
```diff
@@ -11,9 +11,11 @@ services:
       - mongodb
     environment:
       EDDI_VERSION: 4.8
+      EDDI_JVM_OPTIONS: "-verbose:class -Xlog:gc*=debug:stdout:time,uptime,level -XX:+HeapDumpOnOutOfMemoryError -XX:HeapDumpPath=/dumps/oom.bin"
       EDDI_ENV: production
       EDDI_MEMORY_MIN: 1G
       EDDI_MEMORY_MAX: 1G
+      EDDI_JAVA_ENV_TUNE_BUFFER: "jdk.nio.MaxCachedBufferSize=1048576"
```
ginccc: Is this actually used anywhere? if so, where`?

docker-compose.yml
kkorsakov: > this needs to be reverted back to 'latest'

Reverted, thanks.

docker-compose.yml
kkorsakov: Yes, start_eddi.sh has a generic mechanism to pass `-D` system properties as keys prefixed with  EDDI_JAVA_ENV. 
`env_vars=($(printenv | grep EDDI_JAVA_ENV_))` - this code reads them.

So it's used to tune native buffer cache.

PR #234, opened 2020-11-04
git calls bugfix

rolandpickl:
fixed git calls bugs

---
This change is [Reviewable](https://reviewable.io/reviews/labsai/eddi/234)

### DIFF
--- a/gitCallsTask-impl/src/main/java/ai/labs/git/connector/impl/GitCallsTask.java
+++ b/gitCallsTask-impl/src/main/java/ai/labs/git/connector/impl/GitCallsTask.java
@@ -25,6 +25,7 @@ import javax.inject.Inject;
 import java.io.File;
 import java.io.IOException;
 import java.net.URI;
+import java.net.URISyntaxException;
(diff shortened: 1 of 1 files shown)


ginccc:
___
*[gitCallsTask-impl/src/main/java/ai/labs/git/connector/impl/GitCallsTask.java, line 247 at r1](https://reviewable.io/reviews/labsai/eddi/234#-MLI7HKr-Fc86vOlu5Wm:-MLI7HKr-Fc86vOlu5Wn:bejaypf) ([raw file](https://github.com/labsai/eddi/blob/8de1afe6fe83836153d590c1f8df62fab77acd8b/gitCallsTask-impl/src/main/java/ai/labs/git/connector/impl/GitCallsTask.java#L247)):*
> ```Java
>     public void gitPush(GitCall gitCall) {
>         try {
>             Path gitPath = Paths.get(tmpPath + "/" + this.repositoryLocalDirectory);
> ```

I need the templating here, so it can be dynamic

PR #305, opened 2022-08-26
Add found input value to expression to match

rolandpickl:


ginccc:
Thanks for this contribution @rolandpickl! This idea was actually repurposed for audit trail and debugging and incorporated it into the v6.0.0 codebase. Closing this PR since the underlying code has been completely restructured, but your input helped shape the direction. Cheers! 🙏

PR #384, opened 2025-05-12
Resolved mismatch stubbings in PropertySetterTaskTest.java

ARMS2025:
I analyzed the test doubles (mocks) in the test code of the project. In my analysis of the project, I observed that

In the test `executeTask`: 
* the `getLatestData` method for the `currentStep` object:
i) during test execution the method is called with argument `"actions"`, but is not stubbed, resulting in a mismatch stubbing. 

In general, a mismatched stubbing occurs when a method is stubbed with specific arguments in a test but later invoked with different arguments in the code. Mockito recommends addressing this type of issue (https://www.javadoc.io/doc/org.mockito/mockito-core/latest/org.mockito/org/mockito/exceptions/misusing/PotentialStubbingProblem.html).

I propose a solution below to resolve the mismatch stubbing. Happy to modify the pull request based on your feedback.

### DIFF
--- a/src/test/java/ai/labs/eddi/modules/properties/PropertySetterTaskTest.java
+++ b/src/test/java/ai/labs/eddi/modules/properties/PropertySetterTaskTest.java
@@ -88,6 +88,7 @@ public class PropertySetterTaskTest {
         IData<List<Property>> expectedPropertyData = new Data<>("properties:extracted", propertyEntries);
 
         final String propertyExpression = "property(someMeaning(someValue))";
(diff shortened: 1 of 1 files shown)


ginccc:
Hey @ARMS2025, thanks for this contribution last year — the stubbing analysis was spot on at the time!

Since then, the codebase has gone through a major v6.0 rewrite and those stubbing issues were resolved along the way. The tests pass cleanly on main today. Going to close this one as superseded, but thank you for the contribution!

PR #459, opened 2026-04-26
Gemini Embedding & Chroma Support

niedch:
## Summary

Adds support for Gemini embedding models and ChromaDB as a vector store, enabling RAG (Retrieval-Augmented Generation) pipelines with Google's embedding API backed by ChromaDB persistence.
Also adds mise configuration for easy management of java and maven version & adds common task from README file

## Type of Change

- [x] ✨ New feature (non-breaking change that adds functionality)

## Changes Made

- Added `GeminiEmbeddingModel` implementation in `ai.labs.llm` module
- Updated `EmbeddingModelFactory` to instantiate Gemini embeddings
- Added `ChromaEmbeddingStore` implementation in `ai.labs.llm` module
- Updated `EmbeddingStoreFactory` to create ChromaDB clients
- Added `docker-compose.chroma.yml` for local ChromaDB testing
- Added `mise.toml` tasks for common dev commands

## How to Test

1. Create Gemini Embedding Setup
2. Setup Chroma db using `mise chroma`
3. Use for baseUrl `http://localhost:8000`
4. No API Key

## Checklist

- [x] My code follows the project's code style
- [x] I have added tests that prove my fix/feature works
- [x] Existing tests pass locally
- [ ] I have updated documentation if needed
- [x] My commit messages follow conventional commits
- [x] I have not committed any secrets, API keys, or tokens
- [x] This PR has a clear, focused scope

## Summary by CodeRabbit

* **New Features**
  * Added Gemini embeddings provider (configurable model, API key, task type)
  * Added Chroma vector store support for embeddings

* **Chores**
  * Added ChromaDB Docker Compose for local use
  * Added project toolchain pinning and reusable build/dev tasks (including task to launch ChromaDB)
  * Integrated langchain4j-chroma dependency

* **Documentation**
  * Updated RAG docs to reflect expanded provider/store support

* **Tests**
  * Added unit tests covering the new Gemini provider


ginccc (commented):
Great feature addition and nice work, @niedch! implementation looks clean and fits well into the existing architecture. 

Here are a few things that would be worth adding/changing:

* valueOf() in buildGemini() 
Looks like the try-catch was added in e5b3095b and removed again a few mins later in 34da7f2c. Without it, the raw exception leaks the internal class path, and there's no .toUpperCase() so "retrieval_document" (Google's own casing) would throw

* apiVersion is hardcoded to V2
Langchain4j defaults to V1. Since you had it configurable initially (fe55ed34) before hardcoding it (f2a19ed3), maybe keep it as a config param with V2 as default? Would make it work with older Chroma versions too.

* The collection name sanitization 
([^a-z0-9_] → _) could produce a trailing _, which Chroma doesn't allow (edge case, unlikely to happen in practice)

* Docs. rag.md and README.md still show the old provider/store counts (7+5)

* Javadoc on buildGemini()
The other builder methods have Javadoc documenting params and defaults. Would be good to add the same here (default model, 3072 dimensions, taskType default). Also, GoogleAiEmbeddingModel supports outputDimensionality() for Matryoshka truncation. Could be worth exposing as an optional param

Smaller things:
* chroma imports ended up after qdrant, file convention is alphabetical.

* TaskType is imported directly on line 15 but then used fully qualified on lines 105-106. Minor inconsistency.

* mise.toml uses bare mvn while the project uses ./mvnw (pinned to 3.9.12).

--

Would be great if you could address these before merging. 

Thanks for the contribution! 🙏

niedch:
Hi @ginccc Thank you for the Review! I implemented the requested changes, i only left the mise configuration to use mvn because we already specify the maven version as a tool. One question tho do we have some formatting file the specifies the Import ordering?

ginccc:
No, we don't have a formatting rule for import ordering. It was flagged during review but it's purely stylistic, not functionally relevant. No need to change it.

And fair point on the mise/mvn thing, that makes sense 👍

ginccc:
@niedch are you able to finalize this branch in the next 1-2 days so we can merge it? (addressing latest AI PR feedback + deploying manager builds to this EDDI repo)

Would love to have your improvements being part of the next release!

rolandpickl (approved):

niedch:
@ginccc just finialized the PR, addressed the AI Checks and updated the frontend (also converted in the EDDI-Manager branch the Powershell script to bash).

I think this PR is now good to go!

ginccc:
@niedch fantastic work, this feature is a really great contribution to this project!

I saw that the scanner flagged two more issues (1 high, 1 medium). As this is blocking the merge, please check and decide what to do with it (fix or dismiss) so we can merge this.

ginccc (approved):

niedch:
@ginccc Thank you very much for Review! I missed updating the manage.html with the new js bundle, which is now fixed

PR #587, opened 2026-07-03
feat(llm): multi-model cascade — enterprise hardening

ginccc:
## Summary

Full enterprise pass over the **multi-model cascading** feature. A code review found that two documented capabilities didn't exist end-to-end (SSE events, judge model), the audit trail recorded the *wrong* model under cascade, and token/cost/metrics were discarded so the cost-savings pitch was unmeasurable. This PR lands all of it — plus two rounds of adversarial review and the resulting fixes.

Design spec: [`planning/model-cascade-enterprise-hardening-plan.md`](planning/model-cascade-enterprise-hardening-plan.md). Docs rewritten: [`docs/model-cascade.md`](docs/model-cascade.md).

## What changed

**Correctness / compliance**
- **Audit** records the real winning `provider/model` (was task-level default), plus cascade cost + token usage.
- **Agent-mode confidence** auto-routes to judge/heuristic (the `{response,confidence}` wrapper can't wrap a tool loop); agent-mode response now streamed to SSE as a chunk.
- **convertToObject + cascade** honors native `jsonMode` and forces a non-wrapper strategy.
- **Global-var / Qute** resolution for step `type` + params (parity with the standard path).
- **Cooperative cancellation** in the agent tool-loop.

**Broken promises made real**
- **SSE** `cascade_step_start` / `cascade_escalation` wired end-to-end (handler default → `ConversationService` sink → `RestAgentEngineStreaming`).
- **judge_model** implemented — real `judgeModel` config block built via `ChatModelRegistry`.
- `strategy: parallel` warns; the false "budget" javadoc removed.

**Enterprise gaps**
- Per-step **token usage + cost** in trace + `responseMetadata`; **Micrometer** metrics under `eddi.llm.cascade.*`.
- Cascade **ceilings**: `maxTotalDurationMs` + dollar `maxCostPerRun` (from configurable per-step pricing).
- **Configure-time validation** — hard-error only for the new pricing/ceiling fields; warnings for legacy conditions (so an upgrade never stops a previously-loading agent from deploying).
- JSON-parse-first + config-driven / language-agnostic **heuristics**; **live streaming** of the final step; `returnBestAcrossSteps`; lazy base-model creation.

**Extras requested during review**
- **Cross-provider credentials** — a step/judge on a different provider that omits its own `apiKey` (and would inherit the wrong one) is flagged at deploy time + documented.

## Review & fixes

Two adversarial review passes (multi-lens reviewers → independent skeptics) surfaced real bugs that are all fixed with regression tests, notably:
- a live-streamed step that timed out mid-stream leaked provider tokens and re-emitted a different response → streamed steps now aren't cancelled mid-flight;
- `returnBestAcrossSteps` clobbering a live-streamed step; the validator hard-failing previously-loadable stored configs; unclamped heuristic scores; an `unescapeJsonString` ordering bug; judge-confidence and single-line-fence regressions.

Two more latent bugs were caught while writing tests (stray-`"confidence"` false positive; judge-null-params NPE).

## Backward compatibility

All new config fields are optional with today's behavior as defaults. Configs without `modelCascade`, and `enabled: false`, are unaffected. `StreamingResponseHandler`'s cascade methods are `default`. Stored-config load behavior is preserved (validator warns rather than hard-failing legacy conditions).

## Testing

New-code coverage ≈ **92% instruction / 79% branch** (JaCoCo); the project aggregate gate (>90% / >80%) is unaffected. Full touched-area unit suite is green locally. The full `@QuarkusTest` IT suite runs in CI.

## Summary of changes

* **New Features**
  * Added multi-model cascade streaming callbacks and new SSE events for cascade step start and escalation.
  * Introduced cascade sequencing/evaluation strategy enums and new configuration controls for confidence, ceilings, pricing, and optional “best across steps”.
* **Bug Fixes**
  * Hardened confidence parsing for structured and judge responses with safer fallbacks and valid payload handling.
  * Improved cascade streaming/escalation correctness around timeouts, cancellation, and best-response retention.
* **Documentation**
  * Expanded cascade and enterprise hardening docs, plus updated changelog entries.
* **Tests**
  * Added extensive validation, enterprise behavior, coverage, and edge-case tests.
* **Refactor/Chores**
  * Improved observability by aggregating token usage for tool-enabled executions and enriching cascade traces/metrics.


src/main/java/ai/labs/eddi/engine/internal/RestAgentEngineStreaming.java
```diff
@@ -89,6 +95,21 @@ public void onToken(String token) {
                             sendEvent(eventSink, sse, "token", token);
                         }
 
+                        @Override
+                        public void onCascadeStepStart(int stepIndex, String modelType, String modelName, int totalSteps) {
+                            sendEvent(eventSink, sse, "cascade_step_start",
+                                    String.format("{\"stepIndex\":%d,\"modelType\":\"%s\",\"modelName\":\"%s\",\"totalSteps\":%d}", stepIndex,
```
niedch: I like to be type safe so i would probably wrap this in a Object and serialize that

src/main/java/ai/labs/eddi/modules/llm/impl/ConfidenceEvaluator.java
```diff
@@ -58,94 +80,169 @@ private ConfidenceEvaluator() {
...
+     *            optional overrides for the heuristic strategy (null = defaults)
      * @return a result containing the actual response text and confidence score
      */
-    static EvaluationResult evaluate(String strategy, String response, ChatModel judgeModel) {
+    static EvaluationResult evaluate(String strategy, String response, ChatModel judgeModel, HeuristicConfig heuristicConfig) {
         if (response == null || response.isBlank()) {
             return new EvaluationResult("", 0.0);
         }
 
         return switch (strategy != null ? strategy.toLowerCase() : "structured_output") {
```
niedch: I noticed that a couple of time already, is there a reason why the strategy should be string and not a enum with the values? At the moment you have to trust the comment on the definition. I think adding a enum for the evaluationStrategry and general CacasdingStrategy would be benefitial.

niedch (commented):
Looks good overall, just one thing which could avoid some confustions in the future!

aisabella-ai (approved):

PR #588, opened 2026-07-03
feat(attachments): complete multimodal attachments (upload → LLM, 1:1 + groups)

ginccc:
## Summary

Implements `planning/multimodal-attachments-completion-plan.md` end-to-end: a user can upload a file (image, PDF, text/docs), have it forwarded appropriately to the LLM in **both 1:1 and group conversations**, on **any** configured provider, and recall it in later turns via a dedicated tool.

## Type of Change

- [ ] 🐛 Bug fix (non-breaking change that fixes an issue)
- [x] ✨ New feature (non-breaking change that adds functionality)
- [ ] 💥 Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] 📝 Documentation update
- [ ] ♻️ Refactoring (no functional changes)
- [ ] 🔧 Chore (dependency updates, CI changes, etc.)

## Related Issue

Not tracked by an issue — scoped and delivered directly from `planning/multimodal-attachments-completion-plan.md`.

## Changes Made

- **Phase 0 — Foundations & bug fixes.** `@JsonIgnore` the base64 payload so it's never persisted (the `transient` keyword didn't stop Jackson); scrub inline base64 from persisted `attachment_*` context copies; shared `AttachmentTextExtractor` (PDFBox extracted out of `PdfReaderTool`); `ModelCapabilityService` (vision/documents/audio/image-URL, conservative model-aware defaults + per-task/deployment overrides); align `max-body-size` above the attachment cap.
- **Phase 1 — Storage unification + secure upload.** Collapsed the two parallel blob-store abstractions onto one `IAttachmentStore` (deletes/GDPR cascades were hitting a *different* store than uploads, so blobs were never actually deleted). Added `getMetadata`/`grantAccess`/single-item `delete` with owner-or-grant authz, UUID ref hardening, per-conversation quotas, the `storageRef` extraction branch (fixes an orphaned-upload bug), and an owner-checked download/delete REST surface.
- **Phase 2 — Unified `AttachmentForwarder`.** Replaces the image-only `MultimodalMessageEnhancer`: uniform per-file/aggregate caps across all sources, capability-gated **hybrid PDF** handling (native `PdfFileContent` vs PDFBox text), universal text inline, provider image-URL normalization, extracts/errors persisted (never silent). Plus per-task `multimodal` overrides and history extract-stitching so later turns retain PDF/text content.
- **Phase 3 — Group parity.** `DiscussRequest` carries attachments; the service materializes them bound to the group conversation, **grants each member conversation** at fan-out (the sole grant-minting site), injects `attachment_*` on the member's first turn, and re-grants down nested groups.
- **Phase 4 — `readAttachment` tool.** On-demand multi-turn recall (list + read, PDF page-targeted); implicit conversation id (never LLM-supplied); auto-added to the toolset when the turn has attachments. Listing/resolution goes through a new grant-aware `listAccessible()` so group members can discover attachments shared with (not owned by) them.
- **Phase 6 (partial) — Ops.** Forwarder metrics + GDPR portability metadata (metadata only, never raw bytes).
- **Hardening pass.** A typed `AttachmentAccessDeniedException` replaces string-matching on error messages for REST 403-vs-404 handling; group attachment materialization preserves URL-based attachments even when no blob store is configured (only inline base64 needs one).

## How to Test

1. `./mvnw clean verify -DskipITs` — 654 tests pass; JaCoCo enforces the repo's >90% instruction / >80% branch gate on every new/changed class. (DB/HTTP integration tests remain CI-only — local JVMs can't open the loopback selector `SafeHttpClient` needs.)
2. **1:1 upload → LLM**: upload an image/PDF/text file via the attachment upload endpoint, then send a message referencing it to an agent using a vision/document-capable model; confirm the response reflects the file's actual content.
3. **Group parity**: start a group conversation with attachments in the `DiscussRequest`; confirm each member conversation can see and reason about the shared file (owned by the group conversation, granted to members).
4. **Multi-turn recall**: in a later turn, confirm the `readAttachment` tool lists a previously uploaded file by name and reads it back (including page-targeted PDF reads), for both a 1:1 conversation and a group member conversation.
5. **Deletion / GDPR**: delete an attachment (or cascade-delete a conversation) and confirm the underlying blob is actually removed from storage, not just unreferenced.

## Checklist

- [x] My code follows the project's [code style](CONTRIBUTING.md#code-style)
- [x] I have added tests that prove my fix/feature works
- [x] Existing tests pass locally (`./mvnw clean verify -DskipITs`)
- [x] I have updated documentation if needed
- [x] My commit messages follow [conventional commits](CONTRIBUTING.md#commit-convention)
- [x] I have not committed any secrets, API keys, or tokens
- [x] This PR has a clear, focused scope (one concern per PR) — all commits deliver one feature: multimodal attachments end-to-end

## Deferred follow-ups

Phase 5 (multipart 1:1 `say` + EDDI-Manager / eddi-chat-ui frontend work, which live in separate repos — the two-step upload→say flow already works today) and the rest of Phase 6 (nightly orphan-blob reaper, cost estimates, `attachmentsForwarded` audit entry). Group attachment transport is JSON inline; a multipart file-part endpoint variant would be a thin follow-up.

## Summary by CodeRabbit

* **New Features**
  * Added group discussion attachments with shared fan-out, plus model capability gating (vision/documents/audio/image-URL) and per-turn/per-file/aggregate forwarding limits.
  * Added conversation-scoped attachment reading via an LLM tool and expanded GDPR exports to include attachment metadata.
  * Introduced REST endpoints to download and delete individual attachments, including forwardable-inline reporting and safe filename handling.
* **Bug Fixes**
  * Improved attachment authorization, error-to-HTTP mapping, skip behavior under limits, and attachment text/extract stitching across turns (including resume scenarios).
* **Documentation**
  * Updated import/FQN style guidance for backend Java logging conventions.


src/main/java/ai/labs/eddi/datastore/postgres/PostgresAttachmentStore.java
```diff
@@ -185,4 +280,59 @@ public List<Attachment> listByConversation(String conversationId) {
...
+                            "Attachment storage quota exceeded for conversation: %d + %d bytes exceeds limit of %d. Delete some attachments first."
+                                    .formatted(totalBytes, incomingBytes, maxTotalBytesPerConversation));
+                }
+            }
+        } catch (SQLException e) {
+            throw new AttachmentStoreException("Failed to check attachment quota", e);
+        }
+    }
+
+    private void authorize(String owner, ResultSet rs, String requester) throws SQLException, AttachmentStoreException {
```
niedch, line 311: I think at some point we should consider adding a general solution for the authorization in oder to avoid duplicating it in multiple places

src/main/java/ai/labs/eddi/engine/internal/GroupConversationService.java
```diff
@@ -89,6 +90,11 @@ public class GroupConversationService implements IGroupConversationService {
     private final NonceCacheService nonceCacheService;
     private final String defaultTenantId;
 
+    // Field-injected so the direct-construction unit tests stay unchanged; used to
+    // materialize and share discussion attachments with member conversations.
+    @jakarta.inject.Inject
```
niedch: Could you use imports for this

src/main/java/ai/labs/eddi/engine/internal/ConversationService.java
```diff
@@ -672,6 +680,16 @@ public AgentConfiguration.UserMemoryConfig getUserMemoryConfig() {
             public String getUserId() {
                 return userId;
             }
+
+            @Override
+            public ai.labs.eddi.engine.attachments.IAttachmentStore getAttachmentStore() {
```
niedch: same here with the imports

niedch (approved):
Great stuff, looks good to me! Just had some nitpicks and maybe an item for the future

src/main/java/ai/labs/eddi/datastore/postgres/PostgresAttachmentStore.java
ginccc: Agreed. Let's tackle this in another PR

PR #615, opened 2026-07-28
feat(keycloak): EDDI-branded login theme

ginccc:
## Why

Users redirected from the Manager or Workforce UI to Keycloak went from EDDI's amber-on-near-black design to stock Keycloak blue-on-white. This makes the login page look like part of EDDI.

## What

A `login` theme named `eddi` — **three text/asset files and no FreeMarker overrides** — plus the wiring to activate and ship it.

```
keycloak/themes/eddi/login/
├── theme.properties
└── resources/
    ├── css/eddi-login.css
    ├── fonts/noto-sans-latin-variable.woff2   (35 KB)
    └── img/{logo_eddi.png, favicon.ico}
```

Wiring: `loginTheme`/`displayName`/`displayNameHtml` on the realm, a bind mount in `docker-compose.auth.yml`, and `install.sh` fetching the assets and applying the theme via the Admin API.

## How it works

Three layers:

1. `kcHtmlClass=login-pf pf-v5-theme-dark` adds PatternFly 5's **built-in dark theme** to `<html>`, flipping every PF component the login pages use — including ones we'd never enumerate (tiles, data lists, helper text, panels).
2. A `:root` block of `--pf-v5-global--*` overrides recolours PF's dark defaults to EDDI's stone/amber palette.
3. A short list of targeted rules covers what tokens can't express: logo, page background, card border, button label, autofill.

We own **zero FreeMarker**, so a Keycloak upgrade can only break this loudly rather than drift silently.

## Three non-obvious things a reviewer should know

**`styles` replaces, it does not append.** Theme properties merge key-by-key with the child winning, so `styles` must name the parent's `css/styles.css` explicitly or all base styling is silently dropped. It resolves up the inheritance chain, so we ship no copy. This pins the parent's stylesheet list as of 26.0 — flagged in `theme.properties` as an upgrade checkpoint.

**`div.kc-logo-text` is dead CSS in `keycloak.v2`.** The rule exists in the parent stylesheet but the *element* is in no template — it's a leftover from the v1 theme. The logo goes on `#kc-header-wrapper` with accessible image replacement, so screen readers still announce the realm display name. (`displayName` feeds `<title>`; `displayNameHtml` feeds the header — they're different fields.)

**PatternFly's dark theme re-points components at a different tier of globals.** The primary button reads `primary-color--300`/`--400`, the card `BackgroundColor--300`, form controls `BackgroundColor--400`. Overriding only the `--100` tier left the Sign In button Keycloak blue. Both tiers are set. Related: `:where()` specificity 0 does *not* mean `:root` always wins — PF also sets component variables on component elements, which beats `:root` by proximity.

## Design passes

**First pass — correct but not designed.** Typography was PatternFly's Red Hat faces (matching neither the Manager's Noto Sans nor the theme's stated intent, because no `font-family` was ever set); inputs shared the card's exact colour; and one card carried three corner radii. Fixed by bundling the Manager's variable font (latin subset, 35 KB for weights 100–900), recessing the fields, and replacing 8/4/3px with a two-step 12px/6px scale driven through `--pf-v5-global--BorderRadius--sm`.

**Second pass — after actually looking at the rendered page.** Everything before this was measured, never seen. Seeing it made the real problem obvious: the page ignored its own logo's design language. The wordmark is hairline, widely tracked, geometric — an instrument face — while the form beneath it was heavy 700 labels, pure-black input voids and a flat slab button on a dead-flat background. Direction taken: *a machined panel lit from a single thin source*.

- **Signature: a hairline amber light-line across the card's top edge**, brightest at centre and fading to nothing at both ends, with a matching ambient bloom behind the wordmark so the logo reads as the light source. One gesture expressed twice rather than several competing effects. Animates in once on load; suppressed under `prefers-reduced-motion`.
- **Depth** — layered radial wash instead of a flat fill; the card gains a 1px inset top highlight, a falloff gradient and a real shadow.
- **The password-field seam** — field and visibility toggle are siblings in an input-group, each with its own box, meeting in a hard seam with mismatched corners. The group is now the field and its children go flat, so the pair reads as one control.
- **Inputs** move from pure black — which drew more attention than the button — to a translucent recess with an inset shadow and an amber focus glow.
- **Type** echoes the wordmark: title to weight 300 with tracking, labels to 12px tracked caps.
- **Deliberately not done:** no gradient on the primary button — it is already the heaviest element and would compete with the signature. It did gain vertical padding, which takes it to a 44px touch target on mobile.

## Verified against a running container

`quay.io/keycloak/keycloak:26.0`, fresh volume, reading computed styles rather than eyeballing.

- Flows: sign-in · failed login · update password · re-authenticate · logout confirm · forgot password · error page — all themed.
- Contrast (WCAG AA needs 4.5:1): input text **18.9:1**, error-page text **17.7:1**, links **10.6:1**, button label **9.2:1**, labels **6.9:1**, field error **6.4:1**. Keyboard focus ring 2px amber at **8.25:1** (past 1.4.11's 3:1).
- All theme assets return 200, including the inherited `css/styles.css`. Clean import log. No horizontal overflow at 1280 or 375.

Robustness checks worth calling out:

- **An empty theme directory** — what a failed `install.sh` asset download produces — yields **HTTP 200 with the built-in theme** and an ERROR in the log. A cosmetic failure can't take authentication down, which is what makes non-fatal theme downloads the right call.
- **The verification protocol was run against that broken state and fails as designed** — the checks aren't vacuous.
- **The `install.sh` realm update was executed**, not just written: 204, sets all three fields, idempotent on re-run.
- **Every selector was audited against the live DOM.** One dead rule found and removed; `.pf-v5-c-login__main-footer-band` was checked and kept (emitted by `login.ftl`, hidden only by `registrationAllowed: false`).

## ⚠️ Existing installations need one extra step

Realm import is **one-shot** — Keycloak skips realms that already exist, so editing `eddi-realm.json` does not reach an existing deployment. `install.sh` now applies `loginTheme` through the Admin API so upgrades pick it up. For local dev the equivalent is:

```
docker compose -f docker-compose.yml -f docker-compose.auth.yml down -v
```

which **destroys `keycloak-data`** — all users and sessions. Fine for the seeded dev realm; not a general upgrade step.

## Scope and known gaps

Login theme only. The **account console, admin console and email templates remain stock** — adding them later means sibling directories reusing the same palette block. Other realms including `master` are unaffected; `loginTheme` is scoped to `eddi`.

**Recorded rather than fixed — WCAG 1.4.11 on the resting input border.** `#27272a` on `#18181b` is 1.19:1 against a 3:1 requirement for UI component boundaries. Closing it needs roughly zinc-500 — a visible departure from the border colour the Manager uses throughout, so it's a CI decision rather than a CSS one. Focus state is unaffected.

**Fixed in passing** (`99b2e61`): `"temporary": true` on an imported credential does **not** create an `UPDATE_PASSWORD` required action in Keycloak 26 — the seeded users log straight in. The header comment in `docker-compose.auth.yml` claimed otherwise, listed only two of the three seeded users, and labelled their roles loosely. It now states the real behaviour and explains *why* the `"temporary"` flag is misleading, so the next reader doesn't "fix" the realm JSON in the wrong direction.

**Verified after review — a latent privilege issue, not introduced here.** The realm's `defaultRole` (`default-roles-eddi`) composites to `eddi-admin`, `eddi-editor` and `eddi-user`. Checked against a fresh import:

- The **seeded users are fine.** `eddi`, `viewer` and `user` each hold only their explicit `realmRoles` — confirmed both via the Admin API and by decoding real access tokens (`viewer` gets `realm_access.roles: ["eddi-viewer"]`). The comment added in `99b2e61` is accurate.
- **Any user created without explicit `realmRoles` becomes an admin.** A user created through the Admin API with no roles came back with `eddi-admin, eddi-editor, eddi-user`. That is exactly what self-registration produces.

Nothing is exposed today because `registrationAllowed: false`. But flipping that one flag — a small, reasonable-looking change — would silently grant every self-registered user full admin. The fix belongs in `eddi-realm.json` (`defaultRole` should composite to `eddi-user` alone), which is a change to the realm's authorization model and deserves its own PR rather than riding along with a login theme.

Also worth stating plainly: everything here is **measured, not observed** — the page has not been looked at by a human yet, so the aesthetic judgement is still open.

## Notes

No Java changed, so `./mvnw compile` was not run. Full design, verified-facts table and measurements are in [`planning/keycloak-eddi-theme.md`](planning/keycloak-eddi-theme.md).

## Summary by CodeRabbit

* **New Features**
  * Added a branded EDDI Keycloak login theme with refreshed colors, typography, wordmark, and styling improvements, including a theme appearance toggle (system/light/dark) with saved preference.
  * Updated realm branding and enabled login i18n with supported locales and a default locale; also applied theming to the admin console login branding when needed.
* **Bug Fixes**
  * Updated the documented development test-user login behavior to reflect no forced password change on first login.
  * Improved install behavior when theme assets are unavailable by warning and falling back to the default theme without partial themes.
  * Enhanced accessibility by improving error field `aria-describedby` wiring and correcting invalid-state signaling.


ginccc:
Worked through every review comment, including the suppressed low-confidence ones. Fixed in `8c98c1b`.

**Functional — the only one that wasn't cosmetic**

`install.sh` theme downloads are now atomic (@coderabbitai). The old code removed just the failed asset, leaving `theme.properties` and the rest of the directory mounted — so Keycloak would resolve the `eddi` theme and render it with its stylesheet, fonts or scripts 404ing. A *missing* theme is safe (verified earlier: HTTP 200 on the built-in theme, `ERROR ... Failed to find LOGIN theme`), but a *partial* one is not, so the distinction mattered. Any cosmetic failure now discards the whole theme directory, and the changelog's "fallback is safe" claim is qualified.

**`format("woff2-variations")` → `format("woff2")`** (@Copilot, both the inline and suppressed note). Correct — an unrecognised format string makes the browser skip the source entirely and fall back to system fonts. It came from the Manager's generated CSS. Verified all four faces still load and are in use after the change; variable axes are unaffected by the format hint.

**Stylelint** — five declaration-spacing fixes applied.

**Stale comments the review exposed, all mine.** The suppressed note about the header claiming `:root` overrides "always win" was a good catch — it directly contradicts F17 in the plan, which I discovered later and never propagated back. Also corrected: a claim that editing the file busts the cache via the version in `url.resourcesPath` (it doesn't — the path survived a container restart; responses do send `Cache-Control: no-cache`), the typography comment still describing a Latin-only 35 KB font, and the light-mode comment still saying there is deliberately no switcher.

**Docs.** Changelog: corrected the asset inventory (eleven resources, not "three"/"four"), the "first and only JavaScript" claim, and the two passages that contradicted later sections on the switcher and on font delivery. Plan: fixed the broken `§8` anchor, added fonts and scripts to the deliverables tree, and replaced the embedded CSS snippet with a pointer to the real file — that copy had already drifted into showing exactly the bug F18 documents, so keeping a second copy in sync isn't worth it.

**One not applied, with evidence**

@Copilot on `docker-compose.auth.yml`: the suggestion was to reword the comment because mounting over `/opt/keycloak/themes` would shadow *Keycloak's built-in theme directory*. That premise is incorrect — I checked inside the running container:

```
$ ls -la /opt/keycloak/themes/
README.md
eddi          <- our mount

$ ls /opt/keycloak/lib/lib/main/ | grep themes
org.keycloak.keycloak-themes-26.0.8.jar
```

Built-in themes ship inside that jar, so parent theme resolution is unaffected either way. Applying the suggested wording would have introduced a factual error. The comment is sharpened with this evidence instead, since the confusion is understandable.

ginccc:
Round two, in `b39566e4`.

**The WCAG 1.4.11 gap is closed.** It needed a token of its own rather than a tweak: the Manager's `#27272a` is 1.19:1 against our card and the recess adds only ~1.11:1, so neither boundary carried the control on its own. `--eddi-field-border` is now zinc-500 on dark and stone-500 on light, measured on the running page:

| | field border vs card | vs field fill |
|---|---|---|
| dark | **3.67:1** | 4.35:1 |
| light | **4.80:1** | 4.40:1 |

The card keeps `--eddi-border` — a panel edge is decoration, and 1.4.11 applies to the visual information that identifies a *control*. Visible consequence: fields read more strongly than the Manager's equivalents do. That is the cost of the criterion, and it is stated in the file rather than left for someone to rediscover.

**New review notes, all three actioned**

@Copilot `install.sh:966` — correct, and the more useful of the two. The realm update forced `internationalizationEnabled`/`supportedLocales`/`defaultLocale` on *every* run, so an operator who had deliberately curated their locale list would have it reset on each upgrade. Branding is still forced; the locale block is now only seeded when the operator has not enabled internationalisation themselves. Verified both paths against the shipped code:

```
fresh realm (no i18n)                    -> loginTheme=eddi  i18n=True  locales=30  default=en
operator curated (i18n on, de+en only)   -> loginTheme=eddi  i18n=True  locales=2   default=de
```

@Copilot `install.sh:974` — also correct. `curl -sf` exits non-zero on 4xx/5xx, so the real status was lost to the `|| theme_status="000"` fallback and every failure reported HTTP 000. Dropped `-f` on that call. Note the pre-existing client-update call directly above has the same pattern; I left it alone as it is outside this change, but it is worth a follow-up.

@Copilot `docs/changelog.md:71` — fixed; that bullet still described the single-locale workaround as the final state.

@coderabbitai on the plan's `install.sh` snippet — fixed, though not by making the snippet atomic. The current code does not stage to a temp file, and does not need to: any failed theme resource now discards the whole `keycloak/themes/eddi` directory, which makes per-file atomicity moot. Your other point was right and was the more important half — the snippet was stale. Rather than re-sync a copy that had already drifted once, I replaced it with the shipped semantics in prose: realm JSON fatal, theme resources non-fatal but all-or-nothing, and why a *missing* theme is safe (HTTP 200 on the built-in) while a *partial* one is not.

**Still open from round one, unchanged:** the two `docker-compose.auth.yml` and changelog threads GitHub still shows as open are both marked outdated — one was fixed in `8c98c1b`, the other is the built-in-themes-directory claim I declined with container output in the [previous comment](https://github.com/labsai/EDDI/pull/615#issuecomment-5108215076).

ginccc:
Both current review comments addressed — `0d85d434` and `369a1b78`.

**@coderabbitai — duplicate `--eddi-chevron`.** Correct, and worth looking at *why* rather than just deleting the second line. The patch that introduced the chevron matched the same anchor twice, so it put **both copies in the media query and none in `:root[data-eddi-theme="light"]`**. The duplicate was the harmless symptom; the real defect was that choosing "light" explicitly from the toggle while the OS is dark left the chevron at the dark grey, because it fell back to the `:root` value. Duplicate removed, missing declaration added, all three paths verified in the browser — system follows the OS, explicit light `#78716c`, explicit dark `#a1a1aa`.

**@Copilot — `defaultRole` grants admin.** Confirmed and fixed. The composite is now `eddi-user` alone in both places it is declared.

Verified on a **fresh import**, since editing the file does not touch a running realm:

```
default-roles-eddi -> eddi-user, manage-account, offline_access, uma_authorization, view-profile
new user, no roles -> eddi-user + built-ins        (previously: eddi-admin, eddi-editor, eddi-user)
eddi / viewer / user -> admin+editor / viewer / user   (unchanged)
```

Two things worth stating explicitly:

- **Keycloak's built-ins survive** the change — `offline_access`, `uma_authorization` and the account client roles are still in the composite, so refresh tokens and account access are unaffected. That was the main risk of narrowing it.
- **The seeded users were never affected.** They declare `realmRoles` explicitly, and Keycloak's import does not add the default role on top — so the exposure was only ever for users created *without* roles, i.e. self-registration.

**Existing deployments are deliberately not migrated.** Realm import is one-shot, and unlike the branding fields this is an authorization change, so `install.sh` does not force it — an operator may have granted elevated defaults on purpose. The changelog documents the manual step (admin console → Realm roles → `default-roles-eddi` → Associated roles, or `DELETE /admin/realms/eddi/roles-by-id/{id}/composites`).

The two remaining open threads are the outdated ones from round one: the changelog wording fixed in `8c98c1b`, and the `docker-compose.auth.yml` "built-in theme directory" claim I declined with container output [here](https://github.com/labsai/EDDI/pull/615#issuecomment-5108215076).

aisabella-ai (approved):

PR #617, opened 2026-07-28
fix(security): access control, A2A ownership, GDPR & audit ledger (wave 2a)

ginccc:
**Stacked on [#616](https://github.com/labsai/EDDI/pull/616)** (wave 1). Targets `fix/code-review-findings`, so this diff shows only wave 2a. Kept to 66 files — CodeRabbit skipped #616 for being 51 files over its 100-file limit.

Wave 2 of the 124-finding external review, split in two. This half is access control and compliance; the LLM/memory half is a separate stacked PR.

## The pattern

EDDI already has a working guard triad — `OwnershipValidator`, `ConversationAccessGuard`, `HitlAccessGuard` — used correctly in `engine/internal`, `engine/hitl` and the MCP surface. **Every finding here is a place that never called it.** The fixes are mostly one added line; the work was the inventory.

## Sharpest items

**A1 — the SSE turn endpoint had no ownership check, while its non-streaming twin did.**
The turn then ran under the *target* conversation's `userId`: that user's long-term memories loaded into the prompt, tool calls executed in their context. Anyone who learns a `conversationId` gets this — and **A6 hands out `conversationId`s**.

Rather than patch the one endpoint, the guard moved **down into `ConversationService.say`/`sayStreaming`**, before the memory snapshot loads, so no future REST/SSE/MCP/Slack adapter can omit it. The REST layer keeps a check too, so a denial is a plain 403 instead of an error event on an already-200 SSE stream. Two pass-throughs are deliberate and documented in code: a null guard (bean built outside CDI, i.e. unit tests) and `ContextNotActiveException` (Slack drives `say()` on a plain virtual-thread executor with no CDI request context — that path was never checked before and would otherwise start throwing).

**A2 — attachment endpoints authorised the path parameter, not the caller.**
`IAttachmentStore.load(ref, requestingConversationId)` verifies that the *named* conversation owns the blob — and the caller supplies that name. The check was self-satisfying. All five methods now require caller ownership, checked on the request thread *before* the async hop, since `SecurityIdentity` is request-scoped.

**A9 — A2A sat entirely outside the ownership model.** Four separate defects, including conversations created with `userId = null`, which `OwnershipValidator` treats as "legacy — allow" and therefore leaves permanently unowned.

**A17 → G17 — pseudonymisation silently broke the audit ledger's integrity guarantee.** `updateMany($set userId)` with no HMAC recompute, and `userId` is a *signed* field. So **every routine GDPR erasure produced rows cryptographically indistinguishable from tampered ones.** The class javadoc claiming a write-once contract was literally true and substantively false.

**A8 — the inventory was wider than reported.** The review named `/parserstore/parsers`; I enumerated every `@Path`-bearing `IRest*` interface without `@RolesAllowed` and found **six**, including two the review missed entirely (`IRestCapabilityRegistry`, `IRestWorkflowStepStore`). `IRestVersionInfo` was deliberately left alone — it has no `@Path` and is a mixin; reasoning is in the code.

## Judgement calls worth reviewing

- **A3**: `@RolesAllowed` went on the *method*, not the `IRestConversationStore` class. `McpConversationTools` reaches this interface via `RestInterfaceFactory`, which is a real loopback HTTP client that **sends no auth header** — a class-level role would 401 the MCP `list_conversations` tool whenever `authorization.enabled=true`. Read paths are already owner-scoped; the admin role sits only on the one deployment-wide operation with no owner scoping to apply.
- **A5**: guarded rather than deleted. It duplicates `/usermemorystore`, but EDDI-Manager is a separate repo so I could not prove there are no callers.
- **G18**: chose a **per-conversation sequence** over a global hash chain — a global chain would serialise every audit write.

## Docs (I7–I13)

`docs/semantic-parser.md` documented a **stemming extension that does not exist**, four times, including inside the flagship copy-paste config — copying it throws `UnrecognizedExtensionException` and the agent will not start. Its expression table also claimed `number(42)` / `time(15:00)` where the code emits `integer(42)` and epoch millis, so rules written from it never fired. `docs/conversation-memory.md` and `docs/properties.md` both contradicted AGENTS.md §5.1 (`{properties.X.valueString}` fails at runtime — `MemoryItemConverter` puts raw values).

## Verification

- Full unit suite: **12,384 tests, 0 non-environmental failures.** (308 listed failures/errors all carry a loopback/selector/event-loop signature — this machine cannot bind sockets. CI is the gate for those and for `*IT.java`.)
- **Mutation-checked** A1, A2 (and G2, G12, F18 from the sibling PR): revert the fix, confirm a test actually fails, restore. All bite. Verified against **whole test classes** — `-Dtest=Class#method` silently runs 0 tests and exits 0 when the method is in a `@Nested` class, which reads exactly like a pass.

## Summary by CodeRabbit

* **New Features**
  * Added audit ledger integrity verification REST endpoints for conversations and agents.
  * Extended GDPR erasure to cover conversation checkpoints, group transcripts, schedules, plus cache invalidation.
  * Improved peer isolation for task handling and conversation scoping.
* **Security**
  * Added/strengthened role-based access control across administrative/config REST surfaces.
  * Enforced conversation/user ownership checks and improved response/error sanitization to prevent sensitive leakage.
* **Bug Fixes**
  * Tightened retention validation for ended-conversation deletion scheduling.
  * Fixed audit verification/queue handling edge cases.
* **Documentation**
  * Updated Qute template syntax guidance and semantic/pattern-matcher docs.
* **Tests**
  * Expanded authorization, audit verification, error-sanitization, and template-security coverage.


ginccc:
@coderabbitai review

ginccc:
@coderabbitai full review

ginccc:
Thanks — one of these holds and one does not, and I checked both against the running engine rather than reasoning about it.

**Correct, and fixed:** `GlobalVariableResolver`'s javadoc still described the contract as `{{vars.<key>}}` (and said Jinja2). Double braces don't resolve in Qute — they reach the model as literal text — so it was advertising a form that doesn't work.

**Not correct:** dot notation *does* resolve a hyphenated map key. Rendering against the real `TemplatingEngine`:

| template | result |
|---|---|
| `{vars.default-model}` | `claude-sonnet` |
| `{vars['default-model']}` | `claude-sonnet` |

Qute's map resolver takes the whole segment as the key rather than parsing `default-model` as an identifier, so the hyphenated keys in `docs/global-variables.md` are fine as written. Bracket notation works too and is more explicit, but it isn't required.

The expectation was reasonable enough that I turned it into an assertion (`PlaceholderSyntaxContractTest#hyphenatedKeysResolveBothWays`) rather than just replying, so the next reader gets the answer without re-deriving it.

aisabella-ai (approved):

PR #618, opened 2026-07-28
fix(llm): LLM core, persistent memory, migration, import/export (wave 2b)

ginccc:
**Stacked on [#617](https://github.com/labsai/EDDI/pull/617)** → [#616](https://github.com/labsai/EDDI/pull/616). Targets `fix/code-review-access-control`, so this diff shows only wave 2b. 97 files, under CodeRabbit's 100-file limit.

## The tool pipeline had a second door

`executeToolWrapped` is genuinely well-built — one production call site, all seven tool sources routed through it. **Except `McpCallsTask`**, a behaviour-rule-triggered lifecycle task invoking the *same* external MCP tools, which resolved a `ToolExecutor` and called it directly. It also bypassed `ToolApprovalGate`, whose only production caller was `AgentOrchestrator` — so **`hitlConfig.toolApprovals` gated LLM-initiated tool calls and not rule-initiated ones.** A human-approval gate with a hole in it. (F14)

**F18 — `converse_with_agent` had no guardrails at all.** It never consulted `DynamicAgentConfig`, and `allowDelegation` was never checked anywhere in the codebase. Agent A calls B, which calls A — and because no `conversationId` is passed, each hop starts a *fresh* conversation, so the busy-guard never breaks the cycle. Depth and cost grew unbounded. **Prompt injection in a user message was sufficient to start it.**

**F17 — `maxCreatedAgentsPerDiscussion` was enforced per turn, not per discussion.** `sharedCreatedIds` was created fresh in every `buildToolList` call. A 5-member × 3-phase discussion with the default cap of 5 permitted **up to 75 agents deployed to production**.

## Two real bugs the tests caught that the review did not

The review's own F12 fix (cache the workflow traversal that runs 3–4× per LLM task per turn) introduced two defects. They only surfaced because `WorkflowTraversalTest` started returning 0 instead of 1 — and because the triage pass asked *"stale test, bad fixture, or real bug?"* for every failure instead of adjusting tests until green:

1. **The cache memoized failure-derived results.** A traversal whose workflow read threw still cached its empty result for the full TTL and replayed it to the other traversals in the same turn — an agent **silently losing its httpcalls/mcpcalls/RAG configuration**, with nothing in the logs but a single WARN. Worse than the performance problem F12 set out to fix.
2. **The cache key omitted the target class** while the value was cast with an unchecked `(List<StepConfig<T>>)`, justified by a comment asserting a 1:1 mapping that nothing enforces. A future caller asking for the same step type with a different class would get another caller's entry and a `ClassCastException` from a cache hit with no connection to the calling code.

## F13 was "fixed" but unreachable — worth a look

The wave-2 agent added `inheritedParameters` overloads to `SummarizationService` and wrote tests that exercised them **directly**. Those tests passed. But **no caller in `src/main` ever passed the parameter** — `ConversationSummarizer` still called the 4-arg overload. So the rolling summary still could not authenticate and still silently never materialised: the exact bug F13 describes, with green tests over it.

This is the failure mode workstream J is about, and it's why the finding is only closed now that the parameters are threaded `LlmTask → ConversationSummarizer → SummarizationService`, with a test at each hop that fails if they're dropped. Mutation-verified.

⚠️ **`DreamService` deliberately remains on the un-inherited path.** It's a background job with no parent task to inherit from, so it needs a credential source of its own — that's part of wiring Dream up (finding I1) in wave 3. Flagging rather than half-fixing.

## Memory & properties

- **G2 — `scope: "secret"` failed OPEN to plaintext.** On vault failure the method returned the plaintext *before* the scrub block, so the secret persisted **twice**: as a conversation property and as raw `input:initial` data. Vault-disabled is the **default** (`eddi.vault.master-key` ships empty), so this was the common path, not an edge case.
- **G13** — Token-aware windowing could emit a prompt with **no user message at all**. If anchors alone exceeded the budget the code only warned, and the model answered with no idea what was asked.
- **G12** — A turn could be **silently lost**: `replaceOne` with no upsert whose `UpdateResult` was discarded, so a conversation deleted mid-turn by erasure or retention discarded the turn while the caller got a normal response.
- **G5/G6** — `most_accessed` recall did an N+1 write *inside* an open read cursor and was self-reinforcing (only already-top-N entries were incremented, so a new entry could never climb in). And `updatedAt` was refreshed on every longTerm property every turn, so `most_recent` degenerated to "everything is recent" and `deleteOlderThan` never expired anything for an active user.
- **G7** — Re-upserting a recalled entry **silently flipped its owner** to the reading agent (global), or wrote a duplicate row (group).

## Migration

- **B12** — Migration **irrecoverably erased typed BSON values**: it deleted the legacy `value` field unconditionally but only wrote a replacement for String/Map/Integer/Float. Doubles, longs, booleans and arrays were dropped with no error.
- **B13** — The template migrator rewrote **any** `{...+...}` sequence, corrupting JSON bodies and arithmetic that merely sat in a document containing Thymeleaf syntax.
- **B14** — The rename migration skipped when the v6 collection already existed, excluded those documents from the URI rewrite, **and still marked itself complete**.

## Verification

- Full unit suite: **12,384 tests, 0 non-environmental failures**.
- **Mutation-checked**: G2 (twice — a second, sharper mutation removes *only* the input-scrub call, proving both halves are independently covered), G12, and F13's new wiring at the `LlmTask` hop.


ginccc:
@coderabbitai review

ginccc:
@coderabbitai full review

aisabella-ai (approved):

PR #619, opened 2026-07-28
fix(runtime): concurrency, lifecycle, cancellation, graceful shutdown, Dream wiring (wave 3)

ginccc:
**Stacked on [#618](https://github.com/labsai/EDDI/pull/618)** → [#617](https://github.com/labsai/EDDI/pull/617) → [#616](https://github.com/labsai/EDDI/pull/616). 37 files.

**16 fixed, 1 partial.** This is the wave where the obvious fix is often the wrong one, so the reasoning matters more than usual.

## C1 — cancellation cancelled nothing

`CompletableFuture.cancel(true)` does **not** interrupt a `runAsync`/`supplyAsync` body. The JDK documents `mayInterruptIfRunning` as having no effect there. Five call sites in `GroupConversationService` relied on it, so **"cancelled" agent threads kept mutating `gc.getTaskList()`, `gc.getTranscript()` and the errors list after the orchestrator had already persisted the document.**

Replaced with a cooperative `MemberTurnCancellation` token checked at the agent turn's own await points, plus a bounded drain. This is also the root cause of **C7** — `resetStrandedInProgressTasks` iterates a snapshot then mutates by id, so a falsely-"cancelled" thread could flip a task between those two steps and strand the very task the method exists to rescue.

## C4 — a ~100-turn scalability cliff

The inner pipeline was submitted through the **same** bounded pool as the outer coordinator callable, which then blocked on `future.get()`. With no `quarkus.thread-pool.*` overrides that's the 200-thread default: at ~100 concurrent turns every thread is a waiter, no inner task can ever be scheduled, and **every turn fails at the 60s watchdog**. A cliff, not a gradual slowdown.

Nested submissions now route to a virtual-thread executor via a `ThreadLocal` marker scoped to the callable body.

**Why not `CompletableFuture` composition** (the work order's first suggestion): the coordinator's ordering contract is *"the callable returns ⇒ the turn is done"*. Making the outer non-blocking would require an `IEventBus`/`IConversationCoordinator` SPI change **and** would let the next turn of the same conversation start while the previous one is still running — trading a throughput bug for a correctness bug. Virtual threads are safe here: there is not a single `@RequestScoped` bean in `src/main`, and three existing callers (`SlackEventHandler`, `GroupConversationService`, `RestGroupConversation`) already drive this pipeline with no request context on virtual-thread executors. The marker clears before callbacks run, so `submitNext` still schedules on the managed executor exactly as before, and watchdog/timeout semantics are unchanged.

## C9 / C13 — a completed turn could be re-executed

`onComplete` sat inside the `try` whose `catch (Throwable)` called `onFailure` — which resubmitted the **already-executed** callable as a retry. LLM calls, tool side effects and cost, all running twice. Compounded by C13: the coordinator retried 3× blindly, and since `onFailure` can only be raised from inside the executor task, every retry re-ran a turn that may already have spent money.

## C5 — /rerun destroyed output and regenerated nothing

The component-cache **key** was built from a sublist-relative index while the cache **stores** under the absolute index. The output task then ran with `component == null` and no-op'd — *after* the prior output was already deleted. `indexOffset` was already threaded in and used only for HITL bookkeeping.

⚠️ **Why this survived 12,000 tests:** every existing `LifecycleManagerTest` stubs the component map **empty** — the exact condition that hides the bug. The new test populates it at absolute indices and fails if the offset is removed. This is the J4 problem (mocks that erase the defect) in its purest form.

## B2 — the finding's premise was inverted, and I'd flag this one

The review claimed *"interrupt flags swallowed in 18 of 20 handlers"*. Auditing all 28 sites individually found **the opposite**: 14 already restore the flag correctly, 9 rethrow, and only **4** genuinely swallowed it.

The audit found two things the finding didn't:
- **`ScheduleFireExecutor`** — a broad `catch (Exception)` swallowing `InterruptedException` from `latch.await(5, MINUTES)`, so the poller kept firing schedules after being interrupted for shutdown. Invisible to a `catch (InterruptedException)` grep.
- **`NatsConversationCoordinator`** — the *mirror* bug: `interrupt()` called **unconditionally** on `catch (InterruptedException | TimeoutException)`, so a drain timeout left the shutdown thread flagged and would abort the `@PreDestroy` steps running after it.

One restore is deliberately in a `finally` **after** the store round trips rather than at the top of the catch — the sync Mongo driver aborts with `MongoInterruptedException` when the calling thread is flagged, so an early restore would skip the very `EXECUTION_INTERRUPTED` write that branch exists to perform.

## B3 — no graceful shutdown existed at all

`grep -rn ShutdownEvent src/main` matched nothing. A rolling deploy dropped every queued and in-flight turn, with no drain and no readiness flip.

## I1 / G8 — Dream wired up

Per your decision, registered with `ScheduleFireExecutor` rather than deleted, with the ceiling switched from `maxSummarizationCalls` to the dollar-based `maxCostPerRun` the project's own guidance prescribes.

**G8 matters much more now that Dream actually runs:** consolidation upgraded `self` visibility to `global` whenever a group spanned multiple agents — and with `summarizeGroupBy` defaulting to `category` and `preserveAgentProvenance` defaulting to false, cross-agent grouping was the *default* path. Two agents' private memories became one entry every agent could read.

## Partial

**F6** — REST/pipeline half done (`ConnectionCallback` sets cancelled on client disconnect; cancellation checked at more points). The `modules/llm` half — checks inside the tool loop and cascade — is deferred to that workstream rather than reached across into.

## Verification

- Clean compile passed **first attempt**, no repairs needed, despite three cross-workstream signature changes.
- Full suite: **12,633 tests, 0 non-environmental failures.**
- **All four mutation checks bite** — C1, C5, C9, B2 each fail a test when reverted. Verified against whole test classes, and surefire reports were checked to confirm the new test classes actually executed rather than being silently skipped.
- Concurrency tests use latches/barriers to force the interleaving deterministically rather than sleeping, and carry `@Timeout`.

## Summary by CodeRabbit

- **New Features**
  - Added Dream Consolidation scheduling with estimated LLM cost tracking and optional cross-agent maintenance.
  - Added graceful shutdown gating with readiness health signaling and improved draining behavior.
  - Added cooperative cancellation for in-flight group member turns.
  - Added bounded TTL caching for REST semantic parsers.

- **Bug Fixes**
  - Hardened schedule cross-user ownership checks, consistent timezone defaults, and Dream schedule validation.
  - Prevented stale persistence, processing-gauge leaks, stranded queued work, and ensured dead-lettering happens on first failure.
  - Improved REST/SSE resilience for client disconnects and capacity limits; preserved interrupt signals.

- **Documentation**
  - Expanded Dream configuration/scheduling docs and updated changelog.

- **Tests**
  - Added/updated coverage for shutdown, interrupts, cancellation, streaming disconnects, Dream behavior, and schedule authorization.


ginccc:
@coderabbitai review

ginccc:
@coderabbitai full review

aisabella-ai (approved):

PR #622, opened 2026-07-29
feat(operator): foundation for an agent that can safely write

ginccc:
Groundwork for a workspace operator agent that manages a deployment — creating and updating agents and groups — while acting as **the person chatting to it**, with that person's permissions.

**No new capability is granted here.** The operator's endpoint allow-list is untouched and still read-only. What changes is that widening it is now safe, where before it was neither safe nor functional.

## Generated writes did not work at all

`McpApiToolBuilder.buildBodyTemplate` emitted Qute variables for a request body but registered none of them as tool parameters, and `AgentOrchestrator` builds the tool schema from `ApiCall.getParameters()` alone. The model therefore had no documented way to fill a body; with strict rendering off those variables render empty.

**Every generated `POST`/`PUT`/`PATCH` went out structurally valid and semantically empty.** Adding a write endpoint before this fix would have produced garbage requests that fail at the far end rather than at the config.

## The body is now one model-written variable

A per-property template looked more helpful and was worse three ways:

1. Every variable became a **required** tool parameter — `Map<String,String>` has nowhere to record optionality, so a `PATCH` of one field forced the model to restate all the others, turning a partial update into a full overwrite.
2. Values were substituted into the JSON **unescaped** (the templating engine runs in `TEXT` mode), so a value containing a quote could break the body or add fields the schema never declared. For an agent that reads untrusted content and writes through an API, that is an injection boundary.
3. The HITL card shows tool *arguments*, so "the arguments are the request" only holds if the body is one of them.

The shape a decomposed template implied now lives in the parameter description, which names each property with its type and marks which are required.

## Approval patterns can address the endpoint a tool calls

Tool names come from `operationId` or a slug and drift when a spec changes — and `ToolApprovalGate` **allows** an unmatched call, so a renamed or newly generated write arrived ungated and silently. Method and path were available and discarded one line into registration.

```json
{ "requireApproval": ["http.post:*", "http.put:*", "http.patch:*", "http.delete:*"],
  "exempt": ["http.get:*"] }
```

gates every mutation **without naming a single tool**, so a new endpoint cannot arrive ungated. `http.post:/agentstore/agents` addresses exactly one, because different POSTs carry different weight. Both forms speak the same `METHOD /path` vocabulary as the endpoint allow-list, so the two can be generated from one source rather than maintained in two.

**Design rule: enumerate downward, never upward.** Whether something is gated stays in `requireApproval`/`exempt`. A missed exemption costs an approval prompt; a missed requirement is an ungated write. The same reasoning restricts the method qualifier to `http` — the only source whose tools record an endpoint — so `mcp.post:` is rejected at save time instead of saved as a pattern nothing could ever match.

Backward compatibility was the risk worth designing around: making the source itself `http.post` would have stopped `http:*` matching in every existing config, removing gating silently. The source is unchanged; endpoint identity travels in a parallel map behind a `classify()` overload, and a test asserts `http:*` still gates every http tool when no endpoint data is present.

## MCP tool calls can now run as the chatting user

They previously ran as whatever static credential the config named, and a `${caller:token}` there passed through the global-variable and secret resolvers untouched and was sent as the literal placeholder — failing **silently** rather than closed.

The transport supported this all along: `customHeaders` has three overloads and EDDI used the constant one. The per-request `McpHeadersSupplier` overload makes the credential per-call while the client stays cached. The split it produces is the one we want and langchain4j enforces it — `initialize` and `tools/list` carry a null invocation context, so discovery cannot run under one user's credential and then be reused for everyone's calls.

## Fixed in passing: a privilege bug

MCP clients were cached by **URL alone**, so two agents naming the same server with *different* credentials silently shared whichever client was constructed first — the second borrowed the first's authorization. The key now includes a digest of the configured credential: a digest so a literal key never becomes a map key that could reach a heap dump, taken *unresolved* so configs sharing a vault reference still share a client. No per-user client explosion — a caller-bound config yields one client whose supplier reads the caller per request.

## Not solved, deliberately

On an expired MCP session the transport retries `initialize()` on an HTTP callback thread where the caller binding does not exist. That path now sends the request **unauthenticated** rather than falling back to the static key under the caller's intent — a visible failure instead of the wrong authority.

## Verification

**566 tests pass on a clean build** across the affected suites; checkstyle clean. Each fix was mutation-checked by reverting it and confirming a test fails.

Worth flagging for reviewers: **four tests written during this work were found vacuous by that check and rewritten.** Three asserted against a helper or against state seeded through the very wrapper under test; one seeded a thread binding that the fix then restored, so it passed either way. The mutation check caught every one — the tests here are green *and* verified to fail without their fix.

## Review notes

Three earlier claims were corrected during this work, each of which changed a decision:

- `McpCallsConfiguration.toolsWhitelist` does **not** fail closed — filtering is skipped when the list is empty, the same shape as the approval gate.
- `toolApprovals` is **not** agent-only — `LlmConfiguration` carries a per-task override that fully replaces the agent-level block.
- A `PUT` on a config does **not** reach a running agent: `HistorizedResourceStore.update` creates `version + 1` while agents pin a version, so self-modification takes a chain of write → re-point → redeploy, each independently gated. The thing to guard is whatever can re-point a version reference, not `PUT` in general.

## Next

Per-endpoint approval friction (`timeoutPolicy`, `approvalTimeout` and the pause message are still single scalars for every gated tool) · agent-readable documentation (EDDI's MCP client never reads MCP *resources*, so `eddi://docs/*` is reachable from a desktop client but not from an agent) · widening the allow-list · the Manager scope picker and approval surface.

## Summary by CodeRabbit

* **New Features**
  * HITL approvals can now be scoped to HTTP method + normalized endpoint paths (including templated paths).
  * Generated write tools now pass the full request payload as a single `requestBody` parameter.
* **Bug Fixes**
  * Improved MCP discovery/auth isolation, safer cancellation and shutdown, and more reliable retry behavior.
  * Hardened workflow URI parsing and sanitized exception logging to avoid leaking raw messages.
* **Documentation**
  * Updated HITL pattern language plus caller-identity and MCP server guidance, including stricter `${caller:token}` placement rules.


ginccc:
@coderabbitai review

ginccc:
## Review thread disposition

Every unresolved thread checked against current code. GitHub marks a thread *outdated* only when its diff hunk moves, not when the finding is addressed — so several below read as open while already being fixed. Verified individually rather than by that flag.

**Fixed in this round**

| Thread | Status |
|---|---|
| `AgentOrchestrator` — empty configured path yields key `post:` | **Real, fixed** in `d7ea2550e`. `Request.path` defaults to `""` and means the server root; returning it unchanged made a root endpoint ungateable. Mutation-checked. |

**Already addressed in an earlier commit** (verified in the current tree, not assumed)

| Thread | Where |
|---|---|
| `toolCache` keyed by URL only | `ec863045f` — now `toolCache.get(cacheKey(serverConfig))` |
| `validate()` illegal-character message stale | `d763b327f` — message lists `/ { }` |
| `callerBound` misses the bare `{caller:token}` form | `d763b327f` — `rejectUnsupportedReference` runs first |
| `describeBodySchema` says "single JSON object" for arrays | `d763b327f` — branches on `schema.getType()` |
| Caller-token restriction inconsistent across docs | `824ed5be6` — `httpcalls.md` and `security.md` now state query parameters, bodies and paths |
| `withBearerPrefix` returns untrimmed value | `04a20ae9f` — the fix claimed in `ec863045f` had silently failed to apply; corrected and verified in the file |

**Verification:** 816 tests pass on a clean build; checkstyle clean. Each behavioural fix on this branch was mutation-checked by reverting it and confirming a test fails.

**One caveat for the approver.** The central control here — `ToolApprovalGate` — fails *open*: an unmatched pattern leaves the call allowed. Seven review rounds each found a variant of that shape, which is why validation now rejects any pattern that cannot match, checked by compiling the prefix and testing it against the endpoint keys the runtime actually emits rather than by enumerating known-bad forms.

aisabella-ai (approved):

Issue #547, opened 2026-06-18
docs: add Javadoc to secrets package interfaces


ginccc:
## Summary

The `secrets` package implements the Secrets Vault (envelope encryption for API keys, tokens, and other sensitive configuration). Its 3 public interfaces lack Javadoc.

## Interfaces to document (3)

- `src/main/java/ai/labs/eddi/secrets/ISecretProvider.java`
- `src/main/java/ai/labs/eddi/secrets/persistence/ISecretPersistence.java`
- `src/main/java/ai/labs/eddi/secrets/rest/IRestSecretStore.java`

## How to write good Javadoc

1. Read the implementation class to understand what the interface does
2. Look at the model class it operates on
3. Write 2-3 lines: purpose + what consumes it + key behaviors

**Good example** (see `IConversationMemoryStore`):
```java
/**
 * Persistence store for conversation memory snapshots.
 * Called by the lifecycle pipeline to save/load conversation state
 * between turns and across server restarts.
 */
```

**Bad example** (don't do this):
```java
/** Store for agents. */
```

> **Note**: If the interface currently has only `/** @author ginccc */`, replace it with
> meaningful Javadoc. Per CONTRIBUTING.md, `@author` tags are not used in new code.

## Acceptance criteria

- [ ] All 3 interfaces have class-level Javadoc
- [ ] Any existing `@author` tags are removed
- [ ] `./mvnw compile` passes

nightcityblade:
Hi, I'd like to work on this. I'll submit a PR shortly.

nightcityblade:
Quick update: I checked the current main branch and these three interfaces already have class-level Javadoc, so I won't take this one after all.

Issue #545, opened 2026-06-18
docs: add Javadoc to config store interfaces (batch 2 of 2)


ginccc:
## Summary

Continuation of interface Javadoc coverage for the `configs` package.

## Interfaces to document (8)

- `src/main/java/ai/labs/eddi/configs/channels/IChannelIntegrationStore.java`
- `src/main/java/ai/labs/eddi/configs/deployment/IDeploymentStore.java`
- `src/main/java/ai/labs/eddi/configs/descriptors/IDocumentDescriptorStore.java`
- `src/main/java/ai/labs/eddi/configs/propertysetter/IPropertySetterStore.java`
- `src/main/java/ai/labs/eddi/configs/snippets/IPromptSnippetStore.java`
- `src/main/java/ai/labs/eddi/configs/rag/IRagStore.java`
- `src/main/java/ai/labs/eddi/configs/properties/IUserMemoryStore.java`
- `src/main/java/ai/labs/eddi/configs/groups/IAgentGroupStore.java`

## How to write good Javadoc

1. Read the implementation class to understand what the interface does
2. Look at the model class it operates on
3. Write 2-3 lines: purpose + what consumes it + key behaviors

**Good example** (see `IConversationMemoryStore`):
```java
/**
 * Persistence store for conversation memory snapshots.
 * Called by the lifecycle pipeline to save/load conversation state
 * between turns and across server restarts.
 */
```

**Bad example** (don't do this):
```java
/** Store for agents. */
```

> **Note**: If the interface currently has only `/** @author ginccc */`, replace it with
> meaningful Javadoc. Per CONTRIBUTING.md, `@author` tags are not used in new code.

## Acceptance criteria

- [ ] All 8 interfaces have class-level Javadoc (2+ meaningful lines)
- [ ] Any existing `@author` tags are removed
- [ ] `./mvnw compile` passes

AzazelSensei:
Opened https://github.com/labsai/EDDI/pull/740 with class-level Javadoc on the eight interfaces.

Issue #544, opened 2026-06-18
docs: add Javadoc to config store interfaces (batch 1 of 2)


ginccc:
## Summary

Most store interfaces in `configs/` lack Javadoc. They follow a consistent CRUD pattern and can be documented by reading the implementation class.

## Interfaces to document (8)

- `src/main/java/ai/labs/eddi/configs/agents/IAgentStore.java`
- `src/main/java/ai/labs/eddi/configs/apicalls/IApiCallsStore.java`
- `src/main/java/ai/labs/eddi/configs/llm/ILlmStore.java`
- `src/main/java/ai/labs/eddi/configs/mcpcalls/IMcpCallsStore.java`
- `src/main/java/ai/labs/eddi/configs/output/IOutputStore.java`
- `src/main/java/ai/labs/eddi/configs/parser/IParserStore.java`
- `src/main/java/ai/labs/eddi/configs/rules/IRuleSetStore.java`
- `src/main/java/ai/labs/eddi/configs/dictionary/IDictionaryStore.java`

## How to write good Javadoc

1. Read the implementation class to understand what the interface does
2. Look at the model class it operates on
3. Write 2-3 lines: purpose + what consumes it + key behaviors

**Good example** (see `IConversationMemoryStore`):
```java
/**
 * Persistence store for conversation memory snapshots.
 * Called by the lifecycle pipeline to save/load conversation state
 * between turns and across server restarts.
 */
```

**Bad example** (don't do this):
```java
/** Store for agents. */
```

> **Note**: If the interface currently has only `/** @author ginccc */`, replace it with
> meaningful Javadoc. Per CONTRIBUTING.md, `@author` tags are not used in new code.

## Acceptance criteria

- [ ] All 8 interfaces have class-level Javadoc (2+ meaningful lines)
- [ ] Any existing `@author` tags are removed
- [ ] `./mvnw compile` passes

PR #623, opened 2026-07-30
docs: add Javadoc to datastore core interfaces

### LINKED ISSUE #546: docs: add Javadoc to datastore core interfaces
## Summary

The `datastore` package defines the storage abstraction layer that ALL config stores extend. These are the most important interfaces to document because they define the versioned CRUD contract used throughout the entire project.

## Interfaces to document (7)

- `src/main/java/ai/labs/eddi/datastore/IResourceFilter.java`
- `src/main/java/ai/labs/eddi/datastore/IResourceStorage.java`
- `src/main/java/ai/labs/eddi/datastore/IResourceStorageFactory.java`
- `src/main/java/ai/labs/eddi/datastore/IResourceStore.java`
- `src/main/java/ai/labs/eddi/datastore/serialization/IDescriptorStore.java`
- `src/main/java/ai/labs/eddi/datastore/serialization/IDocumentBuilder.java`
- `src/main/java/ai/labs/eddi/datastore/serialization/IJsonSerialization.java`

## How to write good Javadoc

1. Read the implementation class to understand what the interface does
2. Look at the model class it operates on
3. Write 2-3 lines: purpose + what consumes it + key behaviors

**Good example** (see `IConversationMemoryStore`):
```java
/**
 * Persistence store for conversation memory snapshots.
 * Called by the lifecycle pipeline to save/load conversation state
 * between turns and across server restarts.
 */
```

**Bad example** (don't do this):
```java
/** Store for agents. */
```

> **Note**: If the interface currently has only `/** @author ginccc */`, replace it with
> meaningful Javadoc. Per CONTRIBUTING.md, `@author` tags are not used in new code.

## Tips

- `IResourceStore` is the **root interface** â€” start here. All config stores extend it.
- Look at `AbstractResourceStore` to understand the versioning and CRUD lifecycle.
- `IResourceStorage` vs `IResourceStore`: storage is the low-level DB access, store is the higher-level versioned API. Document this distinction.

## Acceptance criteria

- [ ] All 7 interfaces have class-level Javadoc
- [ ] `IResourceStore` Javadoc explains the versioning contract
- [ ] Any existing `@author` tags are removed
- [ ] `./mvnw compile` passes

Kayvan-Zahiri:
Fixes #546

Documents the seven interfaces listed in the issue. Six carried an `@author`-only Javadoc block — the case the issue calls out as needing replacement — and `IResourceStorageFactory` already had prose, so only its `@author` tag is removed.

Each one was written from the interface's own method set plus its implementation (`ResourceFilter`, `PostgresResourceStorage`, `HistorizedResourceStore`/`AbstractResourceStore`, `DescriptorStore`, `DocumentBuilder`, `JsonSerialization`), following the house style already set by `IResourceStorageFactory`.

| Interface | What the doc says |
|---|---|
| `IResourceStore` | The versioned CRUD contract: id + integer version, `update` returns a new version rather than mutating, `delete` marks a version while leaving it readable via `readIncludingDeleted`, `deleteAllPermanently` is the only history-removing call. Also why `IResourceId`, the checked exceptions and `ConfigurationUpdate` are nested in the contract. |
| `IResourceStorage` | The backend-facing half — raw revision persistence without the validation/interception `IResourceStore` layers on top, obtained through the factory rather than injected, and why `MAX_RESULT_LIMIT`/`resolveLimit` live here so backends can't drift. |
| `IResourceFilter` | The query surface on top of id-addressed reads: how `QueryFilters` groups combine via `ConnectingType`, index/limit paging, and that it backs the REST list endpoints. |
| `IDescriptorStore` | Metadata held separately from the configuration document, and that browsing reads descriptors instead of deserializing every document — plus why `NO_LIMIT`/`DEFAULT_LIMIT` exist. |
| `IDocumentBuilder` | The mapping seam between database documents and model classes, so persistence code never hand-rolls it. |
| `IJsonSerialization` | The shared wrapper so mapper settings are declared once rather than per call site. |

## Verification

```
./mvnw compile   →  BUILD SUCCESS
```

Every `{@link}` target was checked to resolve (members against the declaring interface, types against the source tree) — 20 references, none dangling.

Docs only, no behavior change, no `@author` tags left in the seven files.

## One note on the sibling issues

While scoping this I checked #544 and #545 against `main` as well. Their state is mixed rather than open or done:

- **#544** — 7 of 8 still `@author`-only; `IMcpCallsStore` already has prose
- **#545** — only 3 of 8 still need it (`IDeploymentStore`, `IDocumentDescriptorStore`, `IPropertySetterStore`); the other five are already documented

Worth trimming those file lists so the next contributor doesn't rewrite documentation that already exists — that appears to be what happened on #547, where the interfaces turned out to be done. Happy to take either of them next if useful.

## Summary by CodeRabbit

* **Documentation**
  * Clarified versioned resource store behavior, including revision reads, history writing, and how backend limit clamping works.
  * Expanded resource filtering documentation, including how groups are combined during listing/search (within-group AND/OR; groups joined with AND).
  * Improved guidance for store interception configuration, noting the current lack of runtime interceptor behavior.
  * Refined descriptor store, document-to-model mapping, and shared JSON serialization documentation.
  * No functional behavior or public interfaces changed.

### DIFF
--- a/src/main/java/ai/labs/eddi/datastore/IResourceFilter.java
+++ b/src/main/java/ai/labs/eddi/datastore/IResourceFilter.java
@@ -10,9 +10,10 @@ import java.util.List;
  * Query surface for listing resources by field, on top of the id-addressed
  * reads {@link IResourceStore} provides.
  * <p>
- * A caller passes {@link QueryFilters} groups; filters inside a group combine
- * under that group's {@link QueryFilters.ConnectingType} (AND or OR), and the
- * groups themselves are then combined. Paging is index/limit, with the same
+ * A caller passes {@link QueryFilters} groups. Filters inside a group combine
+ * under that group's {@link QueryFilters.ConnectingType} (AND or OR); the
+ * groups themselves are always combined with AND, so a group's connector never
+ * affects how it joins the other groups. Paging is index/limit, with the same
  * ceiling {@link IResourceStorage} enforces.
  * <p>
  * Used by the REST layer to back list endpoints and their search parameters.
--- a/src/main/java/ai/labs/eddi/datastore/IResourceFilter.java
+++ b/src/main/java/ai/labs/eddi/datastore/IResourceFilter.java
@@ -7,7 +7,18 @@ package ai.labs.eddi.datastore;
 import java.util.List;
 
 /**
- * @author ginccc
+ * Query surface for listing resources by field, on top of the id-addressed
+ * reads {@link IResourceStore} provides.
+ * <p>
+ * A caller passes {@link QueryFilters} groups; filters inside a group combine
+ * under that group's {@link QueryFilters.ConnectingType} (AND or OR), and the
+ * groups themselves are then combined. Paging is index/limit, with the same
+ * ceiling {@link IResourceStorage} enforces.
+ * <p>
+ * Used by the REST layer to back list endpoints and their search parameters.
+ *
+ * @param <T>
+ *            the configuration model being queried
  */
 public interface IResourceFilter<T> {
     List<T> readResources(QueryFilters[] queryFilters, Integer index, Integer limit, String... sortTypes)
--- a/src/main/java/ai/labs/eddi/datastore/IResourceStorage.java
+++ b/src/main/java/ai/labs/eddi/datastore/IResourceStorage.java
@@ -10,7 +10,21 @@ import java.util.Collections;
 import java.util.List;
 
 /**
- * @author ginccc
+ * Backend-facing half of the versioned store contract: raw persistence of
+ * resource revisions and their history, without the validation or interception
+ * {@link IResourceStore} adds on top.
+ * <p>
+ * Implementations are database-specific (MongoDB by default, PostgreSQL as an
+ * alternative) and are obtained through {@link IResourceStorageFactory} rather
+ * than injected directly. Reads are version-addressed, {@link #store} appends a
+ * revision, and history is written separately so a superseded revision stays
+ * retrievable.
+ * <p>
+ * {@link #MAX_RESULT_LIMIT} and {@link #resolveLimit} live here so every
+ * backend clamps unbounded queries identically.
+ *
+ * @param <T>
+ *            the configuration model being stored
  */
 public interface IResourceStorage<T> {
 
--- a/src/main/java/ai/labs/eddi/datastore/IResourceStore.java
+++ b/src/main/java/ai/labs/eddi/datastore/IResourceStore.java
@@ -23,7 +23,10 @@ import java.lang.annotation.Target;
  * Nested here because they belong to the contract rather than any one
  * implementation: {@link IResourceId} (the id/version pair), the checked
  * exceptions callers are expected to handle, and {@link ConfigurationUpdate},
- * the interceptor binding that fires when a store mutates a configuration.
+ * which marks the store methods that mutate a stored configuration. It is
+ * declared as an {@link InterceptorBinding}, but nothing implements that
+ * interceptor today, so the annotation has no runtime behaviour: it documents
+ * intent, and callers that need a cache refreshed invalidate it themselves.
  *
  * @param <T>
  *            the configuration model this store persists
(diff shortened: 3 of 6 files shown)


Kayvan-Zahiri:
Good catch, fixed in the latest push.

I checked it against `ResourceFilter.createQuery` before changing anything: each group's `ConnectingType` selects `Filters.and`/`Filters.or` for the filters *inside* that group, and the resulting group expressions are then wrapped in `Filters.and(connectedFilters)`. So the outer join is always AND, and my original "the groups themselves are then combined" left exactly the ambiguity you flagged.

Worded it slightly more explicitly than the suggestion, since the misreading worth preventing is that a group's connector might govern the outer combination:

> Filters inside a group combine under that group's `ConnectingType` (AND or OR); the groups themselves are always combined with AND, so a group's connector never affects how it joins the other groups.

`./mvnw compile` still passes.

Kayvan-Zahiri:
The Secret Scanning failure on this PR isn't from the diff — it's the fork-secrets boundary, and it will hit every external contributor.

The job fails before scanning anything:

```
[labsai] is an organization. License key is required.
🛑 missing gitleaks license. Go grab one at gitleaks.io and store it as a GitHub Secret named GITLEAKS_LICENSE.
```

`GITLEAKS_LICENSE` is empty in the run's env dump. GitHub doesn't expose repository secrets to `pull_request` runs originating from a fork, so `secrets.GITLEAKS_LICENSE` resolves to `""` for outside PRs, and `gitleaks-action` hard-fails on org-owned repos without a license.

Evidence across the last 14 PRs:

| PR | origin | Secret Scanning |
|---|---|---|
| #623 (this one) | fork | **failure** |
| #610-#622 | internal | success ×13 |

Only fork PR, only failure. Everything else here is green — Build & Test, CodeQL, Trivy, Integration Tests, Preflight.

It matters more than one red check, because `gitleaks` is in the `ship` job's `needs:` list (ci.yml:419), so no external contribution can reach a green pipeline while the repo is recruiting through `help wanted` and `good first issue`.

Three options, all maintainer-side:

1. **Run the gitleaks CLI directly instead of the Action.** The scanner itself is MIT-licensed and needs no key — only the Action wrapper requires one for orgs. Keeps fork coverage intact, which the other two lose.
2. **Skip the job for forks**: `if: github.event.pull_request.head.repo.full_name == github.repository`. Honest and explicit, but leaves fork diffs unscanned on the PR (push-to-main still catches them).
3. **Add the `GITLEAKS_LICENSE` secret** — still won't help fork PRs, since the secret isn't exposed to them at all.

Happy to send option 1 as its own PR if you'd like it, though it's a CI change rather than docs so I didn't want to bundle it in here.

Kayvan-Zahiri:
Good catch, fixed in 4b1239f.

I checked before changing it rather than taking the review at face value: no class in `src/main/java` is annotated `@Interceptor` (the single `jakarta.interceptor.Interceptor` import is `DisabledAuthController` using `Interceptor.Priority.LIBRARY_AFTER`, a priority constant), and there is no CDI observer for it either. All 13 non-declaration usages are the annotation sitting on store mutation methods. AGENTS.md says the same thing, which I should have followed.

New wording:

```java
 * exceptions callers are expected to handle, and {@link ConfigurationUpdate},
 * which marks the store methods that mutate a stored configuration. It is
 * declared as an {@link InterceptorBinding}, but nothing implements that
 * interceptor today, so the annotation has no runtime behaviour: it documents
 * intent, and callers that need a cache refreshed invalidate it themselves.
```

I kept the "declared as an `@InterceptorBinding` but nothing implements it" part deliberately. Dropping it and calling it only a marker leaves the next reader to find `@InterceptorBinding` on line 35 and re-derive that nothing fires. The cache claim is the observed behaviour: `RestPromptSnippetStore` calls `snippetService.invalidateCache()` explicitly, and `RestSecretStore` does the same with `secretResolver.invalidateCache(ref)`.

Unrelated to this PR, but the same claim exists elsewhere: `CapabilityRegistryService` line 33 says the registry is rebuilt on agent create/update/delete "via `@ConfigurationUpdate` observer events." There are no such observer events, so that comment describes a mechanism that doesn't run. It's outside #546's scope (datastore interfaces), so I left it alone rather than widen a docs PR. Happy to send it separately if you want it.

`./mvnw compile` passes and javadoc `-Xdoclint:reference` resolves every link in the file.

ginccc (approved):

ginccc:
Thanks for the contribution, looks good to me!
Re fixing secret scan for forks) option 1 would be a good addition and a separate PR is the right approach

PR #624, opened 2026-07-31
fix(ci): run the gitleaks CLI so secret scanning works on fork PRs

Kayvan-Zahiri:
Follow-up to #623, where you said:

> Re fixing secret scan for forks) option 1 would be a good addition and a separate PR is the right approach

This is option 1: run the gitleaks CLI directly instead of the Action.

## The problem

`gitleaks/gitleaks-action` requires a paid `GITLEAKS_LICENSE` for org-owned repositories, and GitHub does not expose repository secrets to `pull_request` runs originating in a fork. The license is therefore absent by design on exactly the runs that need it, and no secret configuration on this side can supply it.

The open pull requests make the pattern clear. Of the 21 open, 12 are from forks: **all 9 that have a Secret Scanning result failed it**, and 3 never ran it. Of the 9 from branches in this repo, 7 passed and 1 has no result. The one same-repo failure (#430) is a Dependabot bump last updated in May whose logs have expired, so I've left it unattributed rather than guess — Dependabot is not the discriminator, since five other Dependabot PRs pass.

Nothing downstream is actually blocked, and I want to be accurate about that: `docker` lists `gitleaks` in `needs:` but is gated on `github.event_name == 'push'`, and `notify-slack` runs `if: always()`. The real cost is that an outside contributor can never get a green run, the check can't be made required without blocking every fork, and `notify-slack` renders a red Secret Scan icon on every fork PR.

## The fix

The scanner is MIT and needs no key — only the Action wrapper is commercially licensed. This installs the CLI the way this workflow **already installs `preflight`**: version and sha256 in the workflow-level `env:` block, hash checked with `sha256sum -c -` before the binary runs. That is a stronger pin than the commit-pinned Action it replaces, which pins the wrapper but not the binary the wrapper fetches at runtime. Findings are redacted with `--redact=100`, so a real hit never prints the secret into a public log.

Removing `GITHUB_TOKEN` loses nothing: the workflow grants `contents: read` only, so the Action could not post PR comments anyway.

## Two decisions worth reviewing

**Scope stays on the event's commits.** A full-history scan on a full-depth checkout reports **61 findings across 16 commits, dated 2016 to 2026**. Most are in files deleted years ago: Keycloak property files from 2016-2019, vendored licence HTML from 2022, an old `keycloak-dev.json`. The licence files are plainly noise, licence prose tripping `generic-api-key`. A few of the old Keycloak entries are UUID-shaped, which is the shape of a real Keycloak client secret, so they may be worth your eye even though the files are long gone and the history is public regardless. I have not reproduced any values, here or anywhere. Either way none of them is something a contributor could fix, so repo-wide scope would fail the build for everyone. This keeps the same scope the Action used, so coverage does not change.

**Correction to my original description.** I first wrote that a full-history scan found 82 findings across 9 commits, concentrated in `docs/gdpr-compliance.md` and test fixtures, and called all 82 false positives. That was measured on a shallow clone, where the graft boundary makes one commit look like it adds entire files, which both inflated the count and misattributed it. On a full-depth checkout, which is what `fetch-depth: 0` gives CI, it is 61 across 16 commits with a different profile, and the blanket "all false positives" reading was not supported by the measurement. The conclusion is unchanged; the numbers behind it were wrong. The commit message on the first commit still carries the old figure, since force-pushing over a published branch seemed worse than leaving a corrected record.

**The scanner's own config now comes from the base branch.** From review feedback, and it turned out to be real. gitleaks reads two files from the checked-out tree: `.gitleaksignore` and `.gitleaks.toml`. On a pull request that tree is the contributor's. I built the attack to be sure: commit a secret, read back the fingerprint gitleaks reports, add that fingerprint to `.gitleaksignore` in the same PR, and the scan passes clean. A permissive `[allowlist] regexes` in an added `.gitleaks.toml` does the same, and this repo has no `.gitleaks.toml` today so adding one meets nothing. (`gitleaks.toml` and `.gitleaks.yaml` are not auto-detected, so those two files are the whole surface.) `--gitleaks-ignore-path` does not fix it: the flag is honoured but the repo-root file is still read alongside it. Restoring both from `base.sha` does, and preserves the base's own entries. A config change now takes effect once merged and reviewed rather than while it is still being proposed.

This was not a regression — `gitleaks-action` read the checked-out config the same way — but fork PRs were never scanned at all before, so there was nothing to bypass. Making fork scanning work is what makes the boundary matter.

**An unresolvable range is an error, not an empty scan.** On a pull request both `base.sha` and `head.sha` must exist in the checkout, or the step exits non-zero with an `::error::` annotation. Narrowing the range and carrying on would convert a broken fetch into a scan that passes without looking at anything, which is the worst failure mode here because it is invisible. Only the push path falls back to the tip commit, and it logs that it did. `fetch-depth: 0` is now load-bearing rather than incidental.

## Verification

Both steps were extracted from the committed YAML and run verbatim against this repo with the real 8.30.1 binary.

| Case | Expected | Result |
| --- | --- | --- |
| PR, valid range, no secrets | pass | exit 0 |
| PR, `base.sha` unresolvable | fail loudly | exit 1, `::error::`, scanner never invoked |
| PR, `head.sha` unresolvable | fail loudly | exit 1, `::error::`, scanner never invoked |
| PR, empty range | pass | 0 commits scanned, exit 0 |
| PR with planted `Authorization: Bearer <40 hex>` | fail | `leaks found: 1`, exit 1 |
| Push, valid `before` | pass | exit 0 |
| Push, `before` all zeros (new branch) | tip commit, logged | exit 0 |
| Push, `before` absent from checkout | tip commit, logged | exit 0 |
| Neither push nor pull request | tip commit, logged | exit 0 |
| Install, correct version | pass | checksum `OK`, ELF x86-64 |
| Install, tampered tarball | reject | exit 1, no binary written |
| Install, wrong pinned hash | reject | exit 1, no binary written |
| Install, nonexistent version tag | fail | exit 22, no binary written |
| Fork allowlists its own secret via `.gitleaksignore` | blocked | exit 1; base's own entries kept |
| Fork adds a permissive `.gitleaks.toml` | blocked | exit 1; file removed and logged |

The detection case is the one that matters. My first attempt used the AWS documentation example key, which gitleaks allowlists — it reported clean. A scanner that detects nothing would have looked correct. The planted-secret case is what proves the job is not vacuous.

This PR is itself from a fork, so the new job runs on it and demonstrates the fix on the exact case it repairs.

Checksum cross-checked against the release's own `checksums.txt`:
`551f6fc83ea457d62a0d98237cbad105af8d557003051f41f3e7ca7b3f2470eb  gitleaks_8.30.1_linux_x64.tar.gz`

`./mvnw compile` clean; changelog updated in the same commit per AGENTS.md.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved secret scanning for pull requests, including contributions from forks.
  * Scans only event-introduced commits and validates available commit ranges.
  * Uses verified scanner versions and preserves configuration across branches.
  * Ensures detected secrets fail CI checks while reducing false positives.
  * Added fallback handling for push events when comparison bases are unavailable.

* **Documentation**
  * Added changelog details covering updated scanning behavior, verification cases, and false-positive handling.


Kayvan-Zahiri:
Thanks — I tested all three rather than assume. One was real and is now fixed, one is real but I don't think it can be safely adopted here yet, and one I'd rather answer than change. Evidence for each.

## 1. Config from the PR tree — real, fixed in 01916d0c4

Confirmed by building it: commit a secret, read back the fingerprint gitleaks reports, add that fingerprint to `.gitleaksignore` in the same PR. Scan passes clean, exit 0. A permissive `[allowlist] regexes` in an added `.gitleaks.toml` does the same, and this repo has no `.gitleaks.toml` today, so adding one meets no resistance.

One correction to the suggestion: `--gitleaks-ignore-path` does **not** fix it. The flag is honoured, but the repo-root file is still read alongside it — pointing `-i` at an empty directory still left the finding suppressed. What works is restoring both files from `base.sha` into the tree before scanning, which is what the new step does. Base entries are preserved; only PR-introduced changes are neutralised.

## 2. `--no-merges` misses merge-resolution secrets — real, but adopting it now would break contributors

You're right about the gap. Repro, secret existing only in the merge commit:

| flags | scanned | result |
|---|---|---|
| `--no-merges` | 2 | no leaks found |
| no flag | 2 | no leaks found |
| `--diff-merges=separate` | 3 | **leaks found: 2** |

The problem is the side effect. On a branch that predates flagged content and then merges `main` in — an ordinary thing to do — `--diff-merges=separate` surfaces **10 pre-existing findings** that the contributor did not introduce and cannot fix. Tested against real history: branch point 2025-12-03, merge current `main` in, range `main..head` gives `no leaks found` under `--no-merges` and `leaks found: 10` under `--diff-merges=separate`.

And it can't be pre-empted by triaging into `.gitleaksignore` first. Fingerprints are `commit:file:rule:line`, and through a merge diff the commit field is the **merge** commit, so the fingerprints are new every time. Overlap between the merge-diff fingerprints and the full-history ones was **0 of 10**. Every future merge that brings in flagged content mints fresh unsuppressable fingerprints.

So the sequencing that actually works is content-based, not fingerprint-based: fix or allowlist the flagged content itself (path/regex allowlist in a committed `.gitleaks.toml`, or `gitleaks:allow` comments), and then `--diff-merges=separate` becomes safe to turn on. That is a change to files I shouldn't be making unilaterally in a CI PR, so I've left the flag alone and documented the gap. Happy to do it as a follow-up if you want it.

Worth noting none of the 12 currently-open fork PRs is old enough to hit this — the oldest is 2026-06-09 and the newest flagged content is 2026-04-15. It's future branches I'd be wary of.

## 3. Push fallback scans only the tip commit — answering rather than changing

Given the triggers are `push: branches: [main], tags: ['[0-9]*']`, the fallback is only reachable on a tag push or a first push to `main`. A tag points at a commit already scanned when it landed on `main`, so the practical gap is small. `github.event.commits` is also capped at 20 entries by GitHub, so it isn't a complete source to fail closed against. The path logs which range it used, so a surprising scope shows up in the job output. I'd rather leave it simple than add machinery for a case that resolves to already-scanned commits — but if you'd prefer it fail closed instead of falling back, that's a one-line change.

## Unrelated correction to my own PR description

My original text said a full-history scan finds 82 findings across 9 commits, all false positives. That was measured on a shallow clone, where the graft boundary makes a commit appear to add whole files. On a full-depth checkout it's **61 across 16 commits, dated 2016 to 2026**, mostly in files deleted years ago. The vendored licence HTML is clearly noise; a few entries in old Keycloak config are UUID-shaped, which is what a real client secret looks like, so they may deserve a look independently of this PR. I haven't reproduced any values. Description is corrected.

Kayvan-Zahiri:
All four valid, all fixed in 99763a2f2.

**`pull_request_target` was dead code** — the workflow triggers on `push` and `pull_request` only, so both arms were unreachable. I'd add that it was worse than dead: a `pull_request_target` branch inside the step that restores config from `base.sha` implies the job was designed for `pull_request_target` semantics, which run with repository secrets against untrusted code. It wasn't, and shouldn't be read as if it were. Removed from the `if:` and the dispatch.

**The stale count was a real miss.** I corrected the changelog and the PR description and left the inline comment carrying the old text. Both the number *and* the attribution were wrong there, since 82-across-9-in-`docs/gdpr-compliance.md` was the shallow-clone artifact. Now reads 61 across 16 commits dated 2016 to 2026, mostly files deleted years ago, matching the changelog.

**Both changelog nits taken.** "Those two" could bind to `gitleaks.toml`/`.gitleaks.yaml` rather than the intended pair, so the sentence now names them and says the negative case was checked rather than assumed. And the table cell now reads "a Bearer token of 40 hex characters" instead of an inline token shape that rendered as a redaction.

Re-verified after the change: 9 scan cases pass, including the planted secret failing the build and both unresolvable-SHA paths erroring loudly rather than narrowing scope.

Kayvan-Zahiri:
Counting nit is fixed in 19c94e9c5 — the table has 14 rows and the summary said 12. The config-bypass row was added late and the sentence above it never caught up.

The critical config finding is the one already fixed in 01916d0, which CodeRabbit marks as addressed. On the other two:

**Merge commits.** I tested the suggested flags rather than assume, and neither form works. Repro: a secret present only in a merge-conflict resolution, in neither parent.

| `--log-opts` | commits scanned | result |
|---|---|---|
| `--no-merges` (current) | 2 | no leaks found |
| `--diff-merges=separate --no-merges` (as suggested) | 2 | **no leaks found** |
| `--diff-merges=combined` | 2 | **no leaks found** |
| `--diff-merges=dense-combined` / `-c` / `--cc` | 2 | **no leaks found** |
| `--diff-merges=separate` | 3 | `leaks found: 2` |

Two things there worth flagging.

`--no-merges` is a documented synonym for `--diff-merges=off`, and later flags win, so `--diff-merges=separate --no-merges` silently cancels itself. `git log --oneline` on the same range gives 2 commits with the pair and 22 with `--diff-merges=separate` alone. Adopting the suggestion verbatim would look like a fix and change nothing.

The combined family is worse than useless here: gitleaks exits 0 and reports clean without scanning the merge, no error, which is a fail-open that reads as a pass. That matches gitleaks#1028, the parser limitation in the citation CodeRabbit surfaced.

So `--diff-merges=separate` alone is the only mode that catches it, and that is the one I measured in the earlier comment: on a branch that predates flagged content and merges `main` in, it surfaces 10 pre-existing findings the contributor did not introduce, and fingerprints from a merge diff carry the merge commit's SHA so `.gitleaksignore` cannot pre-empt them (0 of 10 overlap). Position unchanged: the gap is real, documented, and the fix has to be content-based allowlisting first.

**Fallback scanning only the tip.** Answered in the previous comment. `github.event.commits` is capped at 20 by GitHub so it cannot be a fail-closed source either, and with `push: branches: [main], tags: ['[0-9]*']` the fallback is reachable only on a tag push or a first push, where the commit was already scanned when it landed on `main`. Happy to make it fail closed instead if a maintainer prefers that trade.

Also rebased onto current `main` (merge, not rebase, per AGENTS.md) after #625 landed. Secret Scanning is green on the merged state.

ginccc (approved):

ginccc:
Thank you for your contribution, @Kayvan-Zahiri!

PR #626, opened 2026-08-01
refactor: Wave R — decompose GroupConversationService, AgentOrchestrator and ConversationService

ginccc:
## Summary

**Wave R of `planning/group-collaboration-improvements-plan.md` is complete.** All three god-classes are decomposed, ahead of the ~18 feature items the plan lands on top of them.

| Class | Before | After |
| --- | --- | --- |
| `GroupConversationService` | 4,417 | **1,422** |
| `AgentOrchestrator` | 2,725 | **1,163** |
| `ConversationService` | 2,735 | **1,174** |

**Checkstyle went 8 violations → 6, and every remaining one is a pre-existing `LineLength` in a file this branch never touched. There is no `FileLength` violation left anywhere in the codebase.**

Every extraction is a behaviour-preserving pure move. The three classes remain the facades (`IGroupConversationService` / `IConversationService` unchanged) and delegate. Many methods stay as thin declared delegators specifically because characterization tests reach them via `getDeclaredMethod(...)`, which resolves by exact parameter types — removing them breaks the safety net even though nothing in production would notice. Per plan rule 3.0-4, collaborators are **plain classes constructed by the facade, not CDI beans**, because ~34 test classes construct these services directly.

### R1 — `GroupConversationService` → 8 collaborators

`GroupAttachmentBinder` · `GroupContextBuilder` · `GroupSigningGuard` · `MemberTurnExecutor` · `PhaseExecutionEngine` · `TaskForceEngine` · `GroupHitlCoordinator` · `GroupLifecycleOps`

Plus one **deliberate behaviour change** in its own commit (plan rule 3.0-1): group discussions now participate in graceful shutdown via `rejectIfShuttingDown()` on every entry point that starts, continues or resumes a discussion → `RejectedExecutionException`, already mapped to HTTP 503.

### R2 — `AgentOrchestrator` → the `ToolSourceProvider` SPI

All eight tool sources are providers now, and `buildToolSetup` **iterates them** rather than calling each by name. Adding a tool source is adding a provider — the property that gates Wave 2's I5/I7/I17.

`BuiltinToolsProvider` · `ContextualToolsProvider` · `AttachmentToolsProvider` · `DynamicAgentToolsProvider` · `HttpCallToolsProvider` · `McpToolsProvider` · `A2AToolsProvider`, assembled by `ToolSourceRegistry`, with `ToolContextBudget`, `ToolApprovalGateSupport`, `ToolLoopRunner`, `ToolLoopResumer` and the `IAgentOrchestrator` interface alongside.

Three properties worth calling out:

- **Isolation is structural.** One failing source used to abort assembly for every other — an unreachable MCP server cost the agent its calculator. `ToolSourceRegistry` wraps every `contribute` call, catching `Throwable` (the realistic non-`Exception` is `NoClassDefFoundError` from an optional integration whose dependency is absent at runtime).
- **Merge is deterministic.** First-write-wins per dispatch name, so collisions resolve by provider order rather than map iteration order, and an operator cannot shadow a governed built-in by naming an MCP tool after it.
- **Per-tool provenance survives.** `toolSources` tags come from the tool's own class, not the provider's `source()` — stamping one tag over a contribution that legitimately spans `memory`/`recall`/`builtin` would silently unmatch a `require: ["memory:*"]` approval pattern.

### R3 — `ConversationService` → `ConversationHitlService` + `ConversationStepRunner`

The HITL cluster was the back ~43% and had almost nothing to do with the front. What remained was two things wearing one name: the public `IConversationService` surface, and the machinery that executes a turn.

## Bugs found and fixed along the way

- **Key-rotation-unsafe signature verification** (`GroupSigningGuard`) — resolved a signer's public key via "whichever key is valid right now" instead of the exact version a signature declares, at both call sites. During a rotation overlap window (a scenario the key model explicitly supports) this could self-discard a good signature or verify against the wrong key. The peer-verify cache was also keyed by agent id alone, so a second entry signed with a *different* key version reused the first entry's cached key.
- **Unsigned member contributions** (`MemberTurnExecutor`) — when a group auto-rejected a member's gated tool call, the resulting transcript entry was built without a signature, while the normal path three lines away signed an identically-shaped entry that peers verify identically.
- **Task verification could abort entirely** (`TaskForceEngine`) — a verifier LLM repeating a subject re-verified an already-verified task, and the resulting `IllegalStateException` escaped from a loop sitting outside the enclosing `try`, losing the verifier's transcript entry and its event.
- **A throwing cancel listener left dynamic agents deployed** (`GroupHitlCoordinator`) — `notifyCancelled` ran inside the state-reverting `catch`, so an SSE sink on a closed stream desynced memory from the store, and the cleanup path reads memory.
- **`snippets`/`vars` unprotected** (`HttpCallToolsProvider`, reported by Copilot) — a prompt-injected tool argument could shadow the deployment-config namespace that httpcall templates read.
- **Per-turn dynamic-agent cap** — `maxCreatedAgentsPerDiscussion` started empty on every `buildToolList` call, so a 5-member × 3-phase discussion with the default cap of 5 could deploy up to 75 agents to production.
- Plus an NPE in `GroupSigningGuard`, missing `enableBuiltInTools` gates on two providers, a null-returning `extractText` dereferenced inside a keep-the-turn-alive fallback, and three logging/privacy fixes.

**A graceful-shutdown observer was added and then removed**, not fixed: `GracefulShutdownService` sets `shuttingDown = true` *inside* `drain()`, so an earlier-priority observer runs while the reject gate is still open and races the very drain it was meant to precede.

## Review

Independent multi-agent reviews at four points, plus every automated comment on this PR. All findings are fixed in-branch — see the `fix(review):` commits and changelog entries. Beyond the test suite, the extractions were verified **structurally**: every pre-branch method signature matched against the union of facade + collaborators (none vanished), and a comment-stripped line diff resolved every unmatched statement to an intentional qualifier rewrite.

Two findings were **declined with reasons**: gating `create_sub_agent`/`teardown_agent` at registration (the guardrails are already enforced in the tool bodies, and suppressing registration is a behaviour change inside a pure move), and removing six "unused" parameters (those signatures are pinned by 44 reflective test references that resolve on exact parameter types — the declarations now carry a comment explaining why the obvious fix breaks the build).

## Test plan

- [x] `./mvnw clean compile` — clean
- [x] **2,640 tests** green across every group, orchestrator, conversation and HITL battery
- [x] 12 new focused collaborator test classes, including `ToolSourceProviderTest` (the plan's R2 post-condition: a provider throwing yields an empty contribution and the loop continues)
- [x] `./mvnw formatter:format validate` — clean; 6 pre-existing `LineLength` violations remain, all in untouched files
- [x] Every behavioural fix mutation-checked: reverting each fails exactly its own test
- [x] CI green — Build & Test, Integration Tests, CodeQL, Trivy, Fuzz, Secret Scanning, Preflight

## What's next

Wave 0 foundations (F1–F6) and the feature items (I1–I18). Full sequencing in the plan's §7 dependency graph. Follow-ups recorded as separate tasks rather than folded in here: group-signing replay protection, `requirePeerVerification` being audit-only, whether a moderator should be peer-critiqued, and a `TaskForceEngine` lock-ordering concurrency test.


ginccc:
## Review findings — status

Every open thread on this PR has been addressed. Verified against `HEAD` rather than from memory:

**Fixed (7):**

| Finding | Where | Fix |
| --- | --- | --- |
| Unsigned tool-less contribution on the graceful resume path | `MemberTurnExecutor` | Now signed via `signingGuard.signOutgoingMessage`, same as the normal path three lines above. Two regression tests, mutation-checked. |
| Stale snapshot could abort task verification entirely | `TaskForceEngine` | Both loops re-read live status via `findById`. The escaping `IllegalStateException` came from a loop outside the enclosing `try`, losing the verifier's transcript entry. Three tests. |
| Throwing cancel listener left dynamic agents deployed | `GroupHitlCoordinator` | `notifyCancelled` moved out of the state-reverting `catch`; only the persist may revert now. Two tests. |
| `create_sub_agent` reachable with built-ins disabled | `DynamicAgentToolsProvider` | `enableBuiltInTools` gate added to `contribute()`. |
| Raw model-generated tool arguments written to WARN log | `HttpCallToolsProvider` | Logs tool name and payload length only. |
| `extractText` could return null into a `.length()` call | `TokenCounterFactory` | Never returns null; the `default` arm already established `""` as the empty value. |
| PLAN branch put an empty `members` list behind a "populated by caller" comment | `GroupContextBuilder` | Dead branch replaced with an accurate explanation — no caller ever populated it. |

Separately, `snippets`/`vars` were missing from `RESERVED_TEMPLATE_KEYS` (thanks @Copilot) — a prompt-injected tool argument could shadow the deployment-config namespace httpcall templates read. Fixed, with a test that spells out all eight namespaces so the set cannot drift again.

**Declined, with the reason now recorded in the code (3):**

- **"Useless parameter" on `GroupContextBuilder.selectDefaultTemplate` / `buildPlainTextFallback`** — genuinely unused, and removing them breaks the build. The characterization suite resolves both through `getDeclaredMethod(..., DiscussionPhase.class, List.class, int.class)`, which matches on exact parameter types; 44 test references depend on those signatures.
- **`MemberTurnExecutor.tryResolveMemberToolPause`'s `input`** — passed positionally among nine arguments; dropping one from the middle of that call is the change most likely to be silently mis-applied.
- **`PhaseExecutionEngine`'s `config`** — the four phase executors share one signature so `executeDiscussion` can dispatch on `TurnOrder` without per-branch argument lists, and `executeParallelPhase` is reached reflectively.

A static analyser cannot see a reflective call site or a shared-signature contract, so each declaration now carries a comment explaining why — rather than leaving these to be re-reported every run.

**Also declined:** gating `create_sub_agent`/`teardown_agent` at registration rather than at call time. `CreateSubAgentTool` already enforces `isEnabled()`/`isAllowCreation()` in its body and `TeardownAgentTool` only touches agents created during the discussion, so there is no gap — and suppressing registration changes what the model can see, which is a behaviour change inside a pure move.

CI is green: Build & Test, Integration Tests, CodeQL, Trivy, Fuzz, Secret Scanning, Preflight.

aisabella-ai (approved):

PR #648, opened 2026-08-09
fix(agents): null-version undeploy no-op, deploy under a CHM bin lock, EXECUTE wave deadline, protocol defaults

ginccc:
Wave A of a deep review of the Agent / Group Agent surface. The review's structural finding was that the **group feature layer has been reviewed exhaustively while the agent lifecycle layer it stands on has not** — four of these five fixes come from `AgentFactory`, two of them from a single six-line method.

## 1. `undeployAgent(env, agentId, null)` was a silent no-op — ephemeral agents leaked

`AgentFactory.AgentId` keys on `(id, version)`, so `new AgentId(id, null)` equals no key the environment map ever holds. Both dynamic-agent teardown paths pass `null`:

- `GroupLifecycleOps#cleanupEphemeralAgents` — runs after **every** group discussion
- `TeardownAgentTool` — the LLM-facing teardown tool

So `agentEnvironment.remove(...)` and `deployedAgents.remove(...)` both did nothing, while the caller logged `"undeployed agent '%s'"` at INFO. `agentStore.deleteAllPermanently(agentId)` then ran anyway. Net result: **the constructed agent stayed resolvable through `getLatestReadyAgent` after its configuration had been deleted from the store**, and `eddi_agents_deployed` grew monotonically for the lifetime of the process.

A null version now means *every* deployed version of that agent — the only reading that matches what the teardown callers mean. The REST/admin path passes a real version and stays exact, which is why this was never noticed.

## 2. `deployAgent` ran store I/O inside `ConcurrentHashMap.compute`

`compute` holds the bin lock for the key while the mapping function runs, and the function's real work was `agentStoreClientLibrary.getAgent()` — a store read plus full workflow construction. `ConcurrentHashMap` documents that a mapping function must be short and must not touch other mappings of the same map; this held a bin lock across multi-second I/O, and any re-entrant agent resolution during construction would have deadlocked.

The claim is now a `putIfAbsent` of an IN_PROGRESS placeholder (atomic, no long hold) with the load outside the map. Side benefit: the placeholder is now actually **published**, which `compute` never did — the dummy was only ever returned on the failure path — so a concurrent `getAgent()` can observe "deployment in progress" instead of a bare null.

## 3. `deployedAgents` was mutated unsynchronized

Appended under `synchronized (deployedAgents)`, removed with no lock at all, and read by the Micrometer gauge thread — three unordered accesses to a `LinkedList`. Now a `CopyOnWriteArrayList`, with `addIfAbsent` replacing a non-atomic `contains()`-then-`add()`.

## 4. The TASK_FORCE EXECUTE wave gave up before its turns could have

The wave waited `agentTimeoutSeconds × maxTasksPerAgent`, which ignores retries — under `onAgentFailure=RETRY` a member legitimately gets `timeout × (maxRetries + 1)` — and carries no setup grace, so even a one-task no-retry wave could expire while the member was still inside its own budget (a member turn reaches its response wait only after agent lookup, conversation start and attachment grants).

That is exactly what `PARALLEL_BATCH_GRACE_FLOOR_SECONDS` exists to prevent, and both the parallel debate batch and the bid round **in the same file** already sized themselves through `parallelBatchBudgetSeconds`. The wave now does too, via a new extracted `TaskForceEngine#waveBudgetSeconds` so the derivation is assertable without timing a real wave.

## 5. The documented 180s `agentTimeoutSeconds` default was unreachable

`resolveProtocol`'s fallback handed out a literal `60`, and `McpGroupTools.create_group` hard-coded `60` — while the constant's own Javadoc, the four shipped templates and the published table in `docs/group-conversations.md` all said 180 (the value introduced *because* 60 timed out thinking models during synthesis). Since nothing backfills a `protocol` block at save time, a group saved without one — the common shape — ran at 60.

The defaults now live on `ProtocolConfig` (`DEFAULT_AGENT_TIMEOUT_SECONDS`, `DEFAULT_MAX_RETRIES`) as the single source of truth, referenced by the engine, the MCP tool and the follow-up path (`resolveAgentTimeoutSeconds`, which had its own stray `return 60`).

> **⚠️ Behaviour change, deliberate:** groups with no `protocol` block, and MCP-created groups, go from a 60s to a 180s per-turn timeout. This makes the code agree with the documentation rather than the reverse.

## Also

- Removed an unreachable version comparison in `getAllLatestAgents` — it compared the result of `getLatestAgent`, which already returns the highest version, against itself.
- Downgraded `waitForDeploymentCompletion`'s "still IN_PROGRESS" ERROR to DEBUG when no deployment future was registered. With the placeholder now published that state is reachable and ordinary. Wiring the registration properly is a follow-up.

## Testing

- **820 existing tests green** across `AgentFactory*`, `AgentDeploymentManagement*`, `TaskForceEngine*`, `GroupConversationService*`, `McpGroupTools*`, `AgentGroupConfiguration*`, `RestAgentAdministration*`.
- **19 new tests** in three classes.
- The `AgentFactory` regression tests were **mutation-checked**: reverting the null-version branch fails 4 of 7.

## Summary by CodeRabbit

* **Bug Fixes**
  * Undeploying an agent without specifying a version now removes all deployed versions.
  * Agent deployment tracking is more reliable during concurrent deployments, failures, and redeployments.
  * Task-force execution waits now better account for retries, grace periods, and sequential tasks.

* **Improvements**
  * Groups without protocol settings now use a 180-second agent timeout and two retry attempts.
  * Timeout and retry defaults are applied consistently across group creation and execution.
  * Improved lifecycle documentation and reduced unnecessary logging for routine in-progress states.


ginccc:
Thanks — all four addressed in `ba8f566`.

**@coderabbitai (Major) — deployment restoring an undeployed agent.** Correct, and it was a regression this PR introduced: moving the store load out of `ConcurrentHashMap.compute` is exactly what opened the window. Publication is now `replace(id, OUR placeholder, agent)` so an interleaved undeploy (or a competing redeploy) keeps its outcome.

On your second point — *"do not only replace put with replace; coordinate the map and metric tracking transition together"* — I took the stronger option and **deleted the parallel structure**. `deployedAgents` is gone; `eddi_agents_deployed` is now a `Gauge` derived from the environment maps (READY entries only, matching the old semantics). Two structures holding one fact could never be linearized against each other; one structure needs no linearization. That also subsumes the unsynchronized-`LinkedList` fix this PR originally shipped.

Interleaving regression test added as requested (`undeployDuringLoadIsNotOverwritten`): it blocks the store lookup on a latch, undeploys, then releases, and asserts both the registry and the gauge. Mutation-verified — it fails against a `put`.

**@github-advanced-security (CodeQL, 4× log injection).** Fixed: the three claim-path `log.debug(String.format(...))` calls and the deploy-failure `log.error` now route `agentId` through `LogSanitizer.sanitize`, and the debug calls use `debugf` instead of pre-formatting.

**@coderabbitai (Minor) — stale Javadoc.** Fixed: `@param agentTimeoutSeconds` now documents the 180s default and where the fallback applies.

**@github-code-quality — useless comparison.** Fixed as suggested: `budget > 30` was implied by `budget >= 90`, replaced with `assertNotEquals(30L, budget)` so the "not the old formula's output" intent survives without an always-true condition.

ginccc:
**Final review pass** (independent second-model review of the full diff) — two further items, fixed in the latest commit:

1. **`waitForDeploymentCompletion` armed `orTimeout(60s)` on the *shared* registration future.** `orTimeout` mutates the future it's called on, and `DeploymentListener` hands the same instance to every waiter and to whoever registered the deployment. Pre-#651 the mutation was near-unreachable (only the ZIP importer ever registered); once #651 registers on every REST deploy, one impatient `getAgent` caller would — at its own 60s deadline — complete the shared future exceptionally for **every** other consumer and evict the registration before the real deployment event arrived. Now a timed `get()`, which waits without writing. The newly-reachable `InterruptedException` restores the interrupt flag, and a timeout logs at WARN as *this caller's* outcome rather than ERROR as the deployment's.
2. **Honest gauge semantics.** The gauge comment claimed "matching the old semantics", but one delta is real: the old `deployedAgents` list keyed on (id, version) with no environment, so an agent deployed to both `production` and `test` counted once; counting map entries counts it per environment — per actual deployment. Deliberate and the truthful reading; now stated in the comment and changelog instead of implied away.

The pass also re-verified the deploy/undeploy CAS, the null-version undeploy, and the wave-budget derivation against the live code paths — no further findings.

ginccc:
Fixed the changelog wording nitpick in the latest commit — both bullets now lead with the behavior ("waits with a timed `get()` and no longer mutates the shared registration future"; "counts READY deployments per environment") and the sentence fragment is gone.

aisabella-ai (approved):

PR #649, opened 2026-08-09
fix(agents): dynamic-agent guardrails — permissive fallback on resume, per-member caps, duplicate recruits, V7

ginccc:
Wave B of the Agent / Group Agent review (Wave A is #648). These guard the highest-blast-radius capability in the product — an LLM deploying agents to production — and the review found them to be the weakest-enforced things in the system.

## 1. 🔴 A group's `dynamicAgents` policy silently reverted to fully permissive on a resumed member turn

`resolveDynamicAgentConfig` accepted only a **typed** `DynamicAgentConfig` out of the context value. It is the only place in the codebase that does a typed cast on a `Context` value — every other consumer (`GroupCostLedger`, `GroupLifecycleOps`, `AttachmentContextExtractor`, `OpenAiConversationBridge`) handles the deserialized shape.

A `Context` whose value round-trips through the conversation store comes back as a raw `LinkedHashMap` — `ConversationMemoryStore` rebuilds it as `new Context(type, map.get("value"))`. So any turn running against a **reloaded** step missed the `instanceof` and fell through to `createDefaultDynamicConfig()`: creation, recruitment and delegation all **on**, for a group that may have disabled every one of them.

The trigger is an ordinary group path, not an exotic one:

1. A member agent's gated tool call fires inside a group.
2. `MemberTurnExecutor#tryResolveMemberToolPause` auto-rejects it (`system:group`) and resumes the member conversation.
3. `ConversationHitlService` loads the snapshot **from the store**; `Conversation#resume` re-enters the same LlmTask at the same index (`executeLifecycleFromIndex`).
4. `buildToolList` runs again against the reloaded step → permissive default.

The orchestrator is still blocked inside that call, so the discussion is also still live in `LiveDiscussionRegistry` and the group-gated tools are available on the same leg.

Resolution is now **three-state and fails closed**:

| Context key | Meaning | Result |
|---|---|---|
| absent | standalone agent | permissive default (operator opted in via the whitelist) |
| present, readable (typed **or** map) | the group's policy | that policy |
| present, unreadable | a group is in charge, we cannot parse what it said | **every capability off** |

"The operator said something we cannot parse" must never resolve to "the operator said yes to everything".

## 2. `maxCreatedAgentsPerDiscussion` was enforced per member, not per discussion

`seedCreatedAgentIds` has always read a `dynamicCreatedAgentIds` context variable for the discussion-wide total — its own Javadoc said *"Nothing in `src/main` writes it today"*. So the cap bounded each member conversation independently: a 5-member group with the default cap of 5 could deploy **25** agents to production, while both the field name and `docs/group-conversations.md` promise 5.

`MemberTurnExecutor` now injects `gc.getCreatedAgentIds()` alongside the policy it already injects per turn. The read side was already written and waiting.

## 3. `RecruitAgentTool` could re-recruit a configured member

The Javadoc claimed *"Configured roster and prior recruits both count"*, but `isAlreadyMember` checked only `recruitedAgentIds`, `dynamicMembers` and `memberConversationIds` — and the last holds an agent only once it has **spoken**. A member whose first turn had not come up yet (any member, during the opening phase) could be "recruited" as a duplicate.

`rosterWithRecruits` de-duplicates so nobody spoke twice, but: the recruitment cap was consumed, a misleading `FACILITATION` entry was written, and `addMemberDisplayName` **overwrote the operator-chosen display name with the raw agent id** — visible in the UI and in `followUpWithMember`'s name resolution.

The tool now receives the configured roster (resolved the same way `ArtifactToolsProvider` resolves its artifact policy), and display-name recording became `putIfAbsent`.

## 4. Teardown never freed a creation slot, and a failed delete orphaned the agent

Two problems in `TeardownAgentTool`:

- `createdAgentIds.remove` ran **before** the delete, so a failed delete left the agent untracked and `cleanupEphemeralAgents` never retried it — config and deployment record orphaned.
- The removal was from a *per-turn* list that `seedCreatedAgentIds` rebuilds by unioning every earlier step, so the id came straight back and the cap counted a non-existent agent forever — defeating the tool's entire purpose.

Teardown now records into `dynamic:torn_down_agent_ids`, which the seed subtracts and `propagateDynamicAgentTracking` applies to the group's own tracking. The tracking removal moved after the successful delete.

## 5. V7 resolved — an omitted whitelist no longer skips the dynamic tools

`docs/langchain.md` states twice that omitting `builtInToolsWhitelist` enables all built-in tools ("(all if not specified)", "Omitting `builtInToolsWhitelist` enables all available built-in tools"), and `BuiltinToolsProvider` implements exactly that for the nine plain beans. This provider alone returned early, so an agent with `enableBuiltInTools=true`, no whitelist and `dynamicAgents.enabled=true` got none of them. Tracked as verify-task V7 in `planning/group-collaboration-improvements-plan.md` §2, which sanctions fixing it as a separate labeled behaviour-change commit.

**Deliberately narrower than "all" in one respect:** the omitted case is honoured only when a group policy governs the turn. `dynamicAgents` is a field on `AgentGroupConfiguration`, so a standalone conversation has no configuration surface on which an operator could have declined — handing it unconfigurable, production-deploying capabilities *because it omitted a list* would be a worse defect than the asymmetry being fixed. Under a group policy the operator **has** that surface (`enabled` / `allowCreation` / `allowRecruitment` / `allowDelegation`), which is what makes "all" safe to mean all.

The existing `AgentOrchestratorBuiltInToolWiringTest` turned out not to pin the no-whitelist case at all, so nothing had to be un-pinned; new coverage pins it in both directions.

## Also

The three dynamic-agent tracking keys moved to `MemoryKeys` — the group layer reads them positionally out of a serialized snapshot, so both sides now name one constant instead of two string literals.

> **⚠️ Behaviour changes, deliberate:**
> - A group whose policy cannot be read now gets **no** dynamic-agent capabilities instead of **all** of them.
> - An agent in a group with `enableBuiltInTools=true` and no whitelist now receives the dynamic-agent tools its group policy permits.

## Testing

- **869 tests green** across `GroupConversationService*`, `GroupLifecycleOps*`, `LlmTask*`, `ConversationHitl*`, `ConversationToolResume*`; **487** across the orchestrator/tool suites.
- **25 new tests** in `DynamicAgentGuardrailResolutionTest` and `DynamicAgentWhitelistAndCapTest`, plus 3 added to `RecruitAgentToolTest`.

## Summary by CodeRabbit

* **New Features**
  * Added safer dynamic-agent guardrails for group discussions.
  * Enforced discussion-wide creation limits and prevented duplicate recruitment, including configured members who have not participated.
  * Added policy-aware tool availability when whitelists are omitted.
  * Improved lifecycle tracking, concurrent updates, teardown handling, and cleanup.

* **Bug Fixes**
  * Policies now fail closed when they cannot be safely resolved.
  * Preserved existing agent display names during recruitment.
  * Failed deletions no longer remove agents from cleanup tracking.

* **Documentation**
  * Added a changelog entry covering the updated guardrails and behavior.


ginccc:
Thanks @coderabbitai — six of seven fixed, one declined with reasoning.

**Fixed**

1. **`teardown_agent` not gated on the policy** (security). Correct, and it was a hole *this PR widened*: with V7 the omitted-whitelist path now reaches that branch, and the branch only checked that the stores were non-null while the tool itself carries no `DynamicAgentConfig`. A disabled — or unreadable, hence fail-closed — policy could still undeploy and permanently delete. Now gated on `dynamicConfig.isEnabled()`.
2. **Unreadable roster failed open.** You're right, and I'd argued myself into the wrong default: failing open silently restores the exact defect the roster check exists to prevent. `configuredMemberIds` now returns `Optional`; `Optional.empty()` means *unavailable* (distinct from a successfully-read empty roster) and **withholds `recruit_agent`** for that turn. Same gate-by-absence discipline as `ArtifactToolsProvider`. This covers both your comments on this point (provider side and tool side).
3. **Stale snapshots resurrecting a torn-down agent.** Real, and cheaper to fix than "heavy lift" suggested — the tombstone only has to live on the `GroupConversation`, which already carries `retainedAgentIds` in exactly that shape. New `tornDownAgentIds`, written *before* the merge by `recordTeardown` and consulted *by* the merge, so ordering no longer decides the outcome.
4. **Non-atomic created-agent merge.** Correct. Compound op now runs under the list's monitor — the same pattern `RecruitAgentTool` already uses for `recruitedAgentIds`, rather than introducing a new lock discipline.
5. **`setMemberDisplayNames` installing a `LinkedHashMap`.** Correct and worse than the PR: it meant *every* reload dropped the guarantee the field declares. Now `ConcurrentHashMap` on both branches.

**Declined — atomic reservation of creation capacity before dispatch**

The analysis is right: parallel member turns each read a stale count and can collectively overshoot. But that is the same accepted-overshoot shape this codebase already documents for the cost ceiling — `GroupCostLedger`: *"an already-dispatched, in-flight turn may still push the total past the ceiling; that overshoot is accepted, not prevented"* — and `MemberTurnExecutor` documents the identical bound for nested-group budgets. Closing it properly needs budget **reservation at dispatch** (each turn claiming a slice up front, plus a decision about returning unspent slices), which is a design change, not a fix, and one that should apply to cost and creation together rather than to one of them here.

What the PR does deliver is the change from *unbounded per member* (5 members × cap 5 = 25 agents against a documented cap of 5) to *bounded with a parallel overshoot*. The residue is recorded in the changelog rather than silently left.

7 new tests in `DynamicAgentTrackingMergeTest` (tombstone, ordering-independence, a 16-thread concurrent-merge check, and the reload/display-name invariant); the tombstone guard is mutation-verified. 1107 unit tests green.

ginccc:
Round 2 — @coderabbitai you're right, and thanks for re-checking rather than accepting round 1.

**Fixed: teardown and tracking merges now share one monitor.** Ordering the writes inside `recordTeardown` was not sufficient on its own — I'd closed the write-ordering half and left the check-then-act half open. A merge could read the tombstone set, find the id absent, be descheduled while a concurrent teardown recorded it, and then complete its own `add`. And you're right that the retained branch had the same shape and sat outside the monitor entirely.

Both now run under a shared `dynamicTrackingMutex` on `GroupConversation` (transient + `@JsonIgnore`, the same pattern the class already uses for `artifactAnnounceMutex`), so check-tombstone-then-add and record-teardown are mutually exclusive rather than merely internally ordered. The merge's synchronized region now covers the retained branch too.

Pinned by two 200-round interleaving tests that race a teardown against a merge on the same id — one for `createdAgentIds`, one for `retainedAgentIds`. Both mutation-verified: dropping the `synchronized` block from `recordTeardown` fails them.

The changelog comment appears to have resolved to an analysis chain with no finding attached — I've made the two test counts in that entry unambiguous (869 for the group-service suites, 228 for the merge-affected suites in this round) in case that was the point. Happy to address it directly if there was something more specific.

ginccc:
**Final review pass** (independent second-model review of the full diff) — one documentation-discipline finding, fixed in the latest commit, plus two verifications worth recording:

- **`tornDownAgentIds` rides schema v4 against the version comment's own rule** ("bump whenever a Wave adds a resume-consumed field"). That's now a *recorded decision* rather than an omission: the field fails soft in every skew direction — a legacy document defaults to an empty set (identity hop), and an older pod re-saving a paused document drops the tombstones, after which the worst outcome is a dead agent re-occupying a cap slot and cleanup retrying a deletion that 404s harmlessly. Contrast `runtimePhases`, whose skew corrupts resume bookmarks. A version bump signals "an old pod must not touch this document"; this field doesn't earn that.
- **Verified: no double-add from the V7 change.** `DynamicAgentToolsProvider` is both registered in `buildToolSetup`'s merger *and* called directly from `collectAllBuiltInTools` — but `collectEnabledTools` has no production caller (kept only so `AgentOrchestratorLocalToolAssemblyTest` can pin that the two paths agree tool-for-tool), so exactly one path runs per turn and both now carry the same semantics.
- **Verified: the stored-map conversion round-trips `LifecyclePolicy`.** The enum carries `@JsonValue`/`@JsonCreator`, which the bare `ObjectMapper` honors, so a legitimate stored policy converts cleanly and fail-closed is reserved for genuinely unreadable ones. Failing closed on unknown fields (a doc written by a newer version) is also the right direction: ignoring them could drop a future *restricting* field.

aisabella-ai (approved):

PR #655, opened 2026-08-09
fix(llm): dynamicAgentConfig guardrails failed open on any turn that did not re-inject them

ginccc:
A systematic sweep of the Context serialization boundary: for every context the system injects, what type does it hold **(a)** on the injecting turn, **(b)** after a store reload, **(c)** after crash recovery, **(d)** on another node?

Nine context keys. Eight are strings, lists or maps and survive unchanged. **`dynamicAgentConfig` is the only typed POJO the system injects — and it is the one that gates the tools that deploy real agents to production.**

| key | injected as | after reload | verdict |
|---|---|---|---|
| `groupId`, `groupConversationId`, `groupDepth`, `delegationDepth` | String | String | OK |
| `groupTranscript`, `schedule`, `attachment_N` | List / Map | List / Map | OK |
| `mcp`, `slack`, `lang`, `channelIntent` | String | String | OK |
| **`dynamicAgentConfig`** | **POJO** | **Map** | **broken** |

## 1. The resolver failed open in two independent ways

`MemberTurnExecutor` injects the group's guardrails on every member turn, and deliberately injects an explicitly **disabled** config when the group configured none. Its comment says exactly why: a member turn that finds no config falls back to the STANDALONE default, which is fully permissive — creation, recruitment and delegation all on.

`resolveDynamicAgentConfig` defeated that defense twice:

1. **Current step only.** Any turn that re-enters without a fresh injection — an HITL tool-approval resume, crash recovery, the group follow-up path — saw nothing.
2. **Typed-object assumption.** `ConversationMemoryStore` and `PostgresConversationMemoryStore` both rebuild a stored context as `new Context(type, map.get("value"))`. After **any** reload the value is a `Map`, so the `instanceof DynamicAgentConfig` failed even when the key *was* present.

Both paths landed on the permissive default.

This was the only sibling resolver that did not already handle the resume case — `ContextualToolsProvider#resolveGroupIds` and `resolveDelegationDepth` both fall back to earlier steps, with comments explaining that a resumed turn re-enters without the original context map. Fixed the same way, plus whole-object Jackson coercion of the map form (not field-by-field — a guardrail added later must not silently read back as its permissive Java default).

## 2. Unresolvable guardrails on a group turn now fail CLOSED

A conversation that demonstrably belongs to a group but whose config cannot be resolved returns a disabled config instead of the permissive default — the discipline `GroupTaskToolsProvider` and `ArtifactToolsProvider` already state.

The group probe requires a context entry that actually **carries a value**, not merely a key. "The key exists" is too weak a claim to strip a standalone agent of tools its designer explicitly whitelisted.

## 3. `GroupLifecycleOps.followUp` never injected the guardrails at all

It injects `groupTranscript` / `groupId` / `groupConversationId` but not `dynamicAgentConfig` — so the defense `MemberTurnExecutor` documents was bypassed on the **entire follow-up path**, with no crash or resume needed. Now injected identically.

## 4. Two disagreeing defaults, and a Javadoc that was not true

`ToolAssemblyContext`'s compact constructor normalizes a null config to a **disabled** one. `DynamicAgentToolsProvider` used a **permissive** one. And `contribute()` ignored `ctx.dynamicAgentConfig()` entirely, re-resolving from memory — despite `AgentOrchestrator#toolAssemblyContext` documenting that the value is resolved once "so two providers resolving it independently could [not] disagree". The inconsistency the Javadoc warns about was already in the code. The provider now consumes the resolved value.

## 5. Withheld group tools are logged

`GroupTaskToolsProvider` / `ArtifactToolsProvider` returned an empty contribution silently. Withholding is correct for a forged or finished discussion id — but it is also what happens if the discussion runs on **another node**, and that failure looked like "the model just lost its tools" with nothing to diagnose from.

## Verified, not changed: the node-affinity invariant

`LiveDiscussionRegistry`'s correctness rests on "a member turn always runs in-process". Re-checked against the implementation rather than the changelog note: `NatsConversationCoordinator.publishAndExecute` publishes only `conversationId.getBytes()` as an ordering marker and then runs the callable through `runtime.submitCallable` **locally**. No JetStream consumer deserializes or executes callables, and the payload could not carry one. A member turn cannot be routed off-node. Invariant holds.

## Testing

377 tests across `DynamicAgentToolsProvider` / `AgentOrchestrator` / `GroupLifecycleOps` / `MemberTurnExecutor` stay green. +7 new in `DynamicAgentConfigContextBoundaryTest`, one nest per observation point (a)–(d).

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved dynamic agent configuration handling across resumed conversations and group follow-ups.
  - Preserved configuration settings when conversations are restored from saved data.
  - Added safer fallback behavior when group configurations are missing or invalid.
  - Ensured earlier conversation context is checked when current configuration is unavailable.

- **Documentation**
  - Added changelog details covering configuration handling, tool access safeguards, and verification results.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -5,6 +5,24 @@
 
 ---
 
+## 🔎 fix(llm): dynamicAgentConfig context boundary — guardrails failed open on every turn that did not re-inject them (2026-08-09)
+
+**Repo:** EDDI (`fix/dynamic-agent-context-boundary`)
+
+Systematic sweep of the Context serialization boundary: for every context the system injects, what type does it hold (a) on the injecting turn, (b) after a store reload, (c) after crash recovery, (d) on another node? Nine context keys; eight are strings, lists or maps and survive unchanged. **`dynamicAgentConfig` is the only typed POJO the system injects — and it is the one that gates the tools that deploy real agents to production.**
+
+1. **`resolveDynamicAgentConfig` failed open in two independent ways.** `MemberTurnExecutor` injects the group's guardrails on every member turn, and deliberately injects an explicitly *disabled* config when the group configured none — its comment says exactly why: a member turn that finds no config falls back to the STANDALONE default, which is fully permissive (creation, recruitment and delegation all on). The resolver defeated that defense twice: it read the **current step only**, so any turn that re-enters without a fresh injection (HITL tool-approval resume, crash recovery, the group follow-up path) saw nothing; and it required a live `DynamicAgentConfig` instance, but `ConversationMemoryStore`/`PostgresConversationMemoryStore` rebuild a stored context as `new Context(type, map)` — so after **any** reload the value is a `Map` and the `instanceof` failed even when the key was present. Both paths landed on the permissive default. Fixed: earlier steps are consulted as a fallback (the shape `ContextualToolsProvider#resolveGroupIds` and `resolveDelegationDepth` already use for exactly the resume case — this was the only sibling that did not), and the map form is coerced back whole-object via Jackson rather than field-by-field, so a guardrail added later cannot silently read back as its permissive Java default.
+2. **Unresolvable guardrails on a group turn now fail CLOSED.** A conversation that demonstrably belongs to a group but whose config cannot be resolved returns a disabled config instead of the permissive default — the discipline `GroupTaskToolsProvider` and `ArtifactToolsProvider` already state. The group probe requires a context entry that actually *carries a value*, not merely a key: "the key exists" is too weak a claim to strip a standalone agent of tools its designer explicitly whitelisted.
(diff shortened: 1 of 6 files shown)


aisabella-ai (approved):

ginccc:
Superseded — closing without merging.

While this branch was in review, `main` moved on: #648 (agent-lifecycle-and-group-deadlines), #649 (dynamic-agent-guardrails), #650 (cadence-claim-expiry) and #651 (deployment-wait-machinery) landed and address the same findings, in places more thoroughly than this PR did. Merging it now would duplicate or regress those.

I re-checked every finding in this PR against current `main` rather than assuming. What was already fixed there is dropped; what was genuinely still missing has been rebuilt on top of current `main` in:

- #660 — idle-conversation sweep age signal + `Period` months bug
- #661 — `create_sub_agent` inheritance, allow-list on the resolved provider, orphan rollback, input bounds
- #662 — vault `allowedAgents` enforcement

No work is lost; the review threads here remain readable for the reasoning.

PR #658, opened 2026-08-09
fix(setup): judge a sub-agent's model against the provider it will actually run on

ginccc:
Closes the residual [#656](https://github.com/labsai/EDDI/pull/656) left documented rather than fixed. **Stacked on `fix/agent-setup-hardening`** — base retargets to `main` once that merges.

## The problem

`allowedModels` maps a provider to the models permitted **for that provider**. But when no provider was named, the check accepted a model appearing in *any* provider's list — and `AgentSetupService.resolveParams` then paired it with the **default** provider.

So this config:

```json
"allowedModels": { "openai": ["gpt-4o-mini"] }
```

accepted `createSubAgent(..., provider=null, model="gpt-4o-mini")` and built an **anthropic** agent running `gpt-4o-mini`. That combination fails at model load, and it is not something the operator — who restricted only openai — ever authorised.

## The fix

A model can only ever be paired with one provider, so it is judged against that provider:

- **Provider named or inherited** — unchanged, including the documented *"absent or empty list means no restriction for that provider"*. An explicit per-provider policy that says nothing about this provider is a deliberate silence, and #656's review confirmed keeping that semantic.
- **Provider omitted** — the model is about to land on the default provider, one the caller never chose and the policy never mentions. Refused, with a message naming the providers that *are* covered, instead of silently building an unloadable agent.

## Behaviour change, stated plainly

`createSubAgent_modelAllowed_providerOmitted_positiveCase` previously asserted that a model from another provider's list is accepted with no provider named. It now asserts the refusal. I kept the test with the reasoning inline rather than deleting it, so the change of meaning is visible in the diff rather than hidden by a removal.

Two tests added: naming the provider makes the same pair succeed, and a model from another provider's list is refused when the provider *is* named.

## Testing

88 tests in `DynamicAgentToolsTest` green.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -5,6 +5,22 @@
 
 ---
 
+## 🔎 fix(setup): a sub-agent's model is judged against the provider it will actually run on (2026-08-09)
+
+**Repo:** EDDI (`fix/sub-agent-model-provider-pairing`, stacked on `fix/agent-setup-hardening`)
+
+The residual left documented by #656's review round, now closed.
+
+`allowedModels` maps a provider to the models permitted **for that provider**, but with no provider named the check accepted a model appearing in *any* provider's list — and `AgentSetupService` then paired it with the DEFAULT provider. A config restricting openai to `gpt-4o-mini` therefore built an **anthropic** agent running `gpt-4o-mini`: a combination that fails at model load, and one the operator never authorised.
+
+A model can only ever be paired with one provider, so it is now judged against that provider:
+- **Provider named or inherited** — unchanged, including the documented "absent or empty list means no restriction for that provider". An explicit per-provider statement that says nothing about this provider is a deliberate silence.
+- **Provider omitted** — the model is about to land on the default provider, one the caller never chose. If the policy has no entry for it, the call is refused with a message naming the providers that *are* covered, rather than silently building an unloadable agent.
+
+`createSubAgent_modelAllowed_providerOmitted_positiveCase` changes meaning as a result; it is retained with the reasoning inline rather than deleted, and joined by two new tests (naming the provider still succeeds; a model from another provider's list is refused).
+
+---
+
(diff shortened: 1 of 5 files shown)


ginccc:
Retargeted from `fix/agent-setup-hardening` to `main`.

Targeting a non-`main` base meant this PR was getting almost **no** validation: no Build & Test, no CodeQL, no Trivy, and CodeRabbit reported `Review skipped: reviews are disabled for this base branch`. Only Codacy ran, so the PR looked green while never having been built.

Because this branch is stacked on `fix/agent-setup-hardening` ([#656](https://github.com/labsai/EDDI/pull/656)) — it depends on that PR's `requestedProvider` / `resolvedProvider` / `AgentSetupService.DEFAULT_PROVIDER` and will not compile without them — the diff shown here temporarily includes #656's commits. That collapses to just this change once #656 merges.

**Merge #656 first**; this one is the follow-up it explicitly deferred.

ginccc:
@coderabbitai review

aisabella-ai (approved):

ginccc:
Superseded by a rebuild on current `main` — closing without merging.

`main` moved substantially while this was in review (#648–#651 landed). This branch was based on the older `main` and had become `CONFLICTING`, which on this repo means `ci.yml` stops running entirely while CodeQL/Codacy stay green — so its checks were no longer meaningful.

The content of this PR is **not** lost. It was verified as still-missing against current `main` and rebuilt there:

- #661 — `create_sub_agent` inheritance (incl. the `readAgent(id, null)` fix), allow-list on the resolved provider, model/provider pairing, orphan rollback, input bounds (supersedes #656 and #658)
- #662 — vault `allowedAgents` deploy-time enforcement (supersedes #659)

Every review finding raised here — including all of CodeRabbit's — is carried into the replacement PRs along with its test. The discussion here remains the record of the reasoning.

PR #668, opened 2026-08-11
feat: docs for agents on every surface, MCP resource bridge, strict task-level toolApprovals

ginccc:
Four items making EDDI's documentation and MCP surface genuinely usable by agents, plus the security fix that stops a task-level `toolApprovals` from bypassing the agent gate. Driven by the EDDI-Manager Platform Operator work (write-by-default; llmstore writes behind the Manager's gate-guard); each item survived two critical design passes, and the second pass changed the design twice — details below and in `docs/changelog.md`.

## 1. Real `list_docs` / `read_docs` MCP tools (`McpDocTools`)

`docs/mcp-server.md` has long documented `toolsWhitelist: ["read_docs", "list_docs"]` — tools that did not exist. A whitelist that matches nothing exposes nothing, so anyone copying the example got a silently tool-less server. The tools now exist, delegating to the same `DocsService` as REST and the `eddi://docs/*` resources.

Tools *alongside* resources on purpose: agentic MCP clients — EDDI's own `McpToolProviderManager` included — consume `tools/list` and never call `resources/read`, so resources alone reach desktop clients and no agent. Role check mirrors `IRestDocs`' five-role enumeration via a new `McpToolUtils.requireAnyRole` (EDDI has no role hierarchy; a single-role check of `eddi-viewer` would refuse an `eddi-admin`).

## 2. `eddi.docs.enabled` (default `true`)

One switch in `DocsService` disables every docs surface together (REST list/read, MCP resources, MCP tools). Previously the only "off" was pointing `eddi.docs.path` at a nonexistent directory — a hack that reads as misconfiguration in every diagnostic. Honest scope note: low value, ~15 lines, kept because the cost is near-zero and a policy deserves a switch.

## 3. MCP resource bridge (`exposeResources` on `mcpcalls` configs)

Opt-in per config: synthesizes `<name>_list_resources` / `<name>_read_resource` tools so an agent can reach **any** MCP server's resources — the protocol half tool-consuming agents otherwise never see (langchain4j's client has supported `listResources`/`readResource` all along; nothing called them).

Deliberate choices: off by default (a pre-existing config must not silently grow tools); **not** subject to `toolsWhitelist` (that filter governs server-advertised names; this feature carries its own opt-in, and a server must not be able to occupy the synthesized names); construction is purely local, executors dial lazily through the shared credential-keyed client cache (unreachable server = error tool *result*, not a discovery failure); text capped at 64K chars, binary described rather than base64-dumped into context; same static-config rejections as `discoverTools`.

## 4. Strict task-level `toolApprovals` (`eddi.hitl.tool.task-approvals.mode`, default `strict`)

The load-bearing one. A per-task `toolApprovals` **fully replaced** the agent-level gate (the identical ternary in `LlmTask` and `ToolLoopResumer`), so `requireApproval: []` buried among forty fields of an llmstore document was a complete bypass — reviewed as an ordinary config edit, effective as a security change. It is also what forced EDDI-Manager to hard-refuse llmstore writes for its Platform Operator.

Under `strict` (via the shared `TaskToolApprovalsResolver`), a task block can only **strengthen** the agent gate:

- `requireApproval` = union — semantically exact for the gate's any-match OR; neutralizes the `[]` bypass (`[] ∪ agent = agent`)
- `exempt` = agent's list verbatim, task entries ignored — exempt beats require (`ToolApprovalGate` P1), so a task-added exemption is precisely the ungating vector. **The critical pass killed a string-intersection design here**: a task exempting a strict *subset* of the agent's patterns shares no strings with it and would have silently gated every read
- task `AUTO_APPROVE` (scalar or rule) demoted to `WAIT_INDEFINITELY` unless the agent itself grants it — generalizing the existing inherited-`AUTO_APPROVE` demotion
- auto-approval budget may only shrink, **including past an unset agent cap**: an unset value is the runtime default (2), not "no cap", so a task stating 10 clamps to `min(task, agent ?? default)`. The default constant moved to `ToolApprovalsConfig` so the runtime and the resolver read one source
- `replace` keeps the pre-6.3.0 wholesale override for designs that deliberately loosen one task

`LlmStore` warns at save time about task `exempt`/`AUTO_APPROVE` that strict mode will not honour — visibility, not rejection: stored configs never brick, `replace` mode still honours them. `LlmTaskCoverageTest.toolApprovals_taskOverrideUsed` deliberately flipped from pinning replace semantics to pinning the strict merge; the full contract lives in `TaskToolApprovalsResolverTest` (17 tests).

## Tests

150 green across the touched areas (17 resolver, 7 `McpDocToolsTest`, 5 `McpResourceBridgeTest`, 5 `RestDocsTest`, 43 `LlmTaskCoverageTest`, gate/rules/provider suites unchanged). The `A2AToolProviderManager*`/`Embedding*` loopback errors in the wider package are the documented sandbox socket limitation (AGENTS.md §Build & Test) — CI is the source of truth there.

## Follow-up trail (not in this PR): operator-over-MCP with full guardrails

Verified facts for the next step, so the path is concrete:

1. **Server side is ready today**: quarkus-mcp-server 1.13.1 supports `@Tool(annotations = @Tool.Annotations(readOnlyHint = true, ...))` — EDDI's 76 tools can declare read/write semantics now.
2. **Client side is the gap**: langchain4j-mcp 1.18.1 drops `annotations` from `tools/list` (no annotation type exists in the client jar), so the gate cannot classify foreign MCP tools by hint without an upstream contribution — or, for EDDI's *own* server, a first-party name→readOnly map threaded into `ToolApprovalGate.classify` the same way `toolEndpoints` already is (giving `mcp` tools the same fail-safe "exempt reads, gate everything else" invariant `http.get:*` provides).
3. **Pinning already holds structurally for MCP**: a gated MCP call's args are frozen in the batch and executed as approved — no pre-request resolution step exists to drift. The gap is preview *quality* (name + redacted args vs method/URI/body), not integrity.
4. **Manager guards need an args-based branch**: `self-guard.ts`/`gate-guard.ts` match `requestPreview.uri` and are blind to MCP calls; they need `source == "mcp"` matching on tool name + parsed arguments.
5. **Provisioning gap**: `setup-api`'s `mcpServerUrls` creates mcpcalls configs with no whitelist — a curated MCP tool subset (e.g. only `apply_agent_changes`, `list_agent_resources`) is not yet provisionable in one call.

The single highest-value composite verb to expose either way is `apply_agent_changes`, which collapses the config→workflow→agent→deploy chain the operator currently walks in four gated steps.

## Summary by CodeRabbit

- **New Features**
  - Added MCP tools for listing and reading documentation pages.
  - Added optional MCP resource bridging for listing and reading remote resources.
  - Added a global setting to enable or disable documentation exposure.

- **Improvements**
  - Task-level HITL tool approvals now support strict merging with agent defaults while retaining replacement mode.
  - Strict mode strengthens approval requirements and enforces auto-approval limits.
  - Added authorization checks and clearer responses for unavailable, invalid, or missing documentation.
  - Updated documentation with configuration guidance and usage details.

---

## Follow-up plan

The operator-over-MCP trail above is written up in full as [`planning/operator-mcp-guardrails-plan.md`](planning/operator-mcp-guardrails-plan.md) — four independently shippable phases, ordered so no phase leaves the operator holding ungated write tools, and written to be executed by someone with no context on this PR.


ginccc:
Thanks — this was a high-value review. Four of five addressed in `2edb562b7`, one declined with reasoning. One of the "minor" findings turned out to be the shadow of a real functional bug.

### 🐛 The tool-count nitpick found a dead feature

Chasing the 74-vs-76 inconsistency led to `McpToolFilter` — a **name whitelist** that `list_docs`/`read_docs` were never added to. `ToolFilter#test` returned `false`, so both tools were invisible to every external MCP client. The feature was dead on arrival, and `McpDocToolsTest` missed it because it calls the methods directly, bypassing the filter.

Fixed, plus the systemic guard: `McpToolFilterCoverageTest` now pins **both** directions by reflection — every `@Tool` is whitelisted (the invisible-tool bug), and every whitelisted name is actually declared (stale entries that read as coverage while protecting nothing). Writing it surfaced a second subtlety worth recording: an omitted `@Tool` name defaults to the sentinel `Tool.ELEMENT_NAME`, **not** to blank, which `McpGroupTools` relies on for all 18 of its tools.

### 🔒 Remote-content boundary — agreed, and it was the sharper version of your point

You're right, and the inconsistency was the damning part: `governDescription` has sanitized remote **tool descriptions** since finding F16, while the resource bridge piped remote **content** to the model unfiltered — the easy way around a guard the tool path already had, over a strictly larger surface.

All remote text now goes through the same `DIRECTIVE_PATTERN` redaction, is delimited as untrusted server data, and is bounded **per field**. Your sub-point about the cap was a genuine bug: the old code appended an unbounded `description` and only checked `sb.length()` afterwards, so one oversized field sailed past the aggregate limit. Lines are now assembled and length-checked *before* appending. Covered by `remoteContentIsGoverned` and `oversizedFieldsAreBounded`.

### 🧪 Test determinism — agreed

The test asserted a property of the CI host, not of this code. `getOrCreateClient` is now a package-visible seam (same precedent as the existing `fetchToolsFromServer`) and the test stubs a throwing `McpClient`.

### 📄 Both docs findings — agreed

`hitl.md`'s Configuration section still stated the pre-6.3.0 "full replace" rule, contradicting the new Precedence row; a reader stopping there would author a task-level `exempt` expecting it to loosen the gate. Corrected to describe the mode-dependent behavior and point at Precedence. `mcp-server.md`'s whitelist paragraph now says 76 and names `McpDocTools`.

### ⏭️ Declined: isolating blocking MCP resource calls onto a bounded executor

Correct that `listResources`/`readResource` are blocking in langchain4j 1.18.1-beta28 — but this is **not specific to the resource bridge**. Every tool source in EDDI executes synchronously on the tool-loop thread: existing MCP tool executors from `McpToolProvider` wrap the equally blocking `client.executeTool`, and `HttpCallToolsProvider`'s executors block on HTTP. Adding a bounded executor for these two synthesized tools alone would be inconsistent with every neighbouring tool and would not fix the underlying property.

If EDDI wants non-blocking tool execution, it belongs at the `ToolExecutor` boundary for all sources, as its own change with its own cancellation and budget semantics — not smuggled in via one feature's two tools. Happy to open a follow-up issue if you'd like it tracked.

### ℹ️ Already fixed before the review landed

The `LlmStore` nitpick (gate the warnings on the configured mode) was fixed in `07949e264`, which post-dates the commit you reviewed — for exactly your reason: in `replace` mode those settings *are* honoured, so warning that they're ignored would be flatly wrong.

aisabella-ai (approved):

ginccc:
All four accepted and fixed in `94a808f5c` — these were design flaws in the **plan**, caught before anyone implemented it, which is exactly what a plan review is for. Two of them would have produced ungated writes.

**1. Annotation values, not presence.** Right, and this was the sharpest one. Phase 2 derives the read-only registry *from* those annotations, so a write tool mistakenly carrying `readOnlyHint = true` passes a presence check and then gets exempted by `mcp.readonly:*` — a mislabelled annotation silently becomes an ungated write. The test now pins an **explicit expected read-only set** with exact match in both directions, so adding a tool forces a deliberate edit to that set — which is the review moment where "is this really read-only?" actually gets asked. Plus a mutating-prefix assertion (`create_|update_|delete_|deploy_|…` must be `readOnlyHint = false`), so a rename into a mutating shape cannot keep a stale hint.

**2. Unauthenticated identity probe.** Agreed, and the plan was plainly wrong here. Probing `/administration/docs` proves only that the server can answer a probe — a hostile one answers it, then exposes a write tool under a read-only name and inherits the exemption. An unauthenticated probe is a liveness check, never a trust decision. The plan now requires the trust decision to be **configuration-side and authenticated** (URL matches the deployment's own configured endpoint *and* the connection carries credentials it issued, or a pinned cert / trusted origin), and anything unproven stays `absent` — which the gate semantics already treat as "gate it".

**3. Redacted/truncated arguments.** Agreed, and it's inconsistent that I didn't apply my own rule. `arguments` is a display payload — redacted and size-capped — so searching it and finding nothing is a **false negative**: a silently unguarded write. Now fails closed on `argsTruncated` **and** on unparseable payloads, matching the `bodyTruncated` rule `gate-guard.ts` already enforces today. Only a complete, parseable payload that demonstrably lacks the id may pass.

**4. Three-tool list as the boundary.** Agreed — and this is the same bug class that already bit this PR: `McpToolFilter`'s name whitelist shipped `list_docs`/`read_docs` invisible, and `tool-scopes.ts` is an allow-list rather than a deny-list for precisely this reason. The MCP branch now **classifies positively**: an MCP tool is a write unless positively known read-only, reusing Phase 2's classification rather than maintaining a second hand-written list. An unrecognised future tool is covered by default.

I also promoted the three generalisable lessons into the plan's guardrails section, so the next reader inherits them instead of rediscovering them:

- never let an enumerated list be the boundary — classify positively
- never derive trust from the untrusted party
- a display payload is not an authority

Thanks — this materially improved the plan.

PR #672, opened 2026-08-11
chore: remove the Agent Father, superseded by the Platform Operator

ginccc:
The Agent Father was EDDI's conversational agent-creation wizard — shipped as a ZIP and deployed on first install. Both of its jobs are now done in EDDI-Manager by the **Platform Operator** (`/manage/operator`) and the form-based **agent wizard** (`/manage/agents/wizard`), both of which call `AgentSetupService` — already the Java equivalent of its 12-step workflow.

It was also already half-broken: the bundled ZIP's entry names use literal backslashes, so `importInitialAgents` 500s on Linux. That is why `src/main/docker/seed-demo-agent.sh` was written to route around it.

## What was removed

| Layer | Change |
|---|---|
| Shipped artifact | `src/main/resources/initial-agents/` (ZIP + `available_agents.txt`) |
| Import machinery | `POST /backup/import/initialAgents` and its implementation |
| Java mentions | Javadoc / tool-description references in `AgentSetupService`, `McpSetupTools` |
| Installers | `install.sh`, `install.ps1`, `gcp/provision-vm.sh` |
| Docs | 3 deep-dive pages + `docs/your-first-agent/` deleted; ~20 mentions rewritten across 11 files |

## Why the import endpoint went too

`available_agents.txt` listed exactly one file — the Agent Father ZIP. With the ZIP gone the endpoint could only ever return an empty list, so leaving it would mean three installers POSTing to a no-op and then printing a banner promising an agent that will not be there.

Removing it also freed `RestImportService` of two injected dependencies it used nowhere else (`IDeploymentListener`, `IRestAgentAdministration`) — constructor narrowed, six unit tests updated. The endpoint is **not** in the operator's allow-list (`EDDI-Manager/src/lib/operator/tool-scopes.ts`), so operator activation is unaffected.

## Why the reference config stayed (renamed)

`docs/agent-configs/agent-father/` → `docs/agent-configs/rule-based-reference/`, descriptors and intro text rebranded. Deleting it would have cost two things unrelated to whether the agent ships:

1. **AGENTS.md §5.6's only worked example** of `actionmatcher`+`inputmatcher`, property setters, httpcalls templating and quick replies.
2. **Real sweep coverage.** It is the only fixture in either root of `StrictBoundaryShippedConfigsTest` / `RuleSetStoreShippedRulesetsTest` supplying a `.agent.json`, `.workflow.json` or `.property.json` — the `src/test/resources/tests` corpus uses the legacy `.bot.json` / `.package.json` names, which `BY_SUFFIX` does not map, so those files count as *unmapped and skipped*.

Post-rename the sweeps still report 32 configs / 7 rulesets checked. §5.6 now states explicitly that it is a fixture, not something that ships.

## `architecture.md` case study: retargeted, not dropped

The point it made — a meta-agent built from ordinary EDDI primitives, self-modifying the system — is *more* true of the Platform Operator, so the section was rewritten around it and now also carries the approval-gate reasoning that makes it safe.

## Second commit: four defects found in a self-review pass

- **`install.sh`** set `JQ_AVAILABLE` for the agent-count check and nothing else, so deleting that check orphaned the detection block. The Keycloak section runs its own `command -v jq` test and is unaffected.
- **`GroupTemplateService`** explained its index file as "the `initial-agents/` pattern" — a Javadoc pointer to a directory this PR deletes. Replaced with the actual reason (a classpath directory cannot be enumerated portably from inside a JAR).
- **`architecture.md` described the operator's gate wrong.** I wrote `requireApproval: ["http:*"]` plus a spec-derived exempt list — which is what `planning/operator-write-scope-plan.md` proposes, and that plan is marked **superseded** at the top. The shipped `buildToolApprovals()` gates by method (`http.post|put|patch|delete:*`, exempting `http.get:*`). Corrected against the code, per AGENTS.md §2 rule 7.
- **An overstated coverage claim** in the first commit message (it named `.httpcalls.json`, which `tests/useCases` also has). Narrowed in the changelog to what is verifiable.

## Deliberately left alone

`docs/changelog.md`, `HANDOFF.md` and `docs/release-notes-6.0.2.md` keep their mentions — dated records of what was true at the time, not live documentation.

## Verification

- `clean compile` + `test-compile` green
- The nine affected test classes green
- `validate` (Checkstyle) clean; `formatter:format` produced no churn
- All four shell/PowerShell scripts parse
- Re-verified: no CI or smoke test touches the removed endpoint; `IDeploymentListener` still has three other consumers; operator activation does **not** require OIDC (`authModeUnusable = oidcEnabled && authMode === "none"`), so pointing a default no-auth install at it is valid; `docs/creating-your-first-agent/` still covers the tutorial gap left by deleting `your-first-agent/`

## Companion change (separate repo, not yet pushed)

`eddi-website` advertised the Agent Father in its Multi-Agent feature list across **11** locales — `ar.ts` had *translated* the name (`الوكيل الأب`), so the obvious grep found only 10. Replaced with the Platform Operator; committed locally on `chore/replace-agent-father-with-platform-operator`.

## Summary by CodeRabbit

* **New Features**
  * Added the Rule-Based Reference Agent as a documented reference configuration.
  * Installation now starts without a pre-deployed starter agent; create agents through the Platform Operator or agent wizard.

* **Documentation**
  * Updated setup, architecture, security, secrets, and integration guidance for the new agent-creation experience.
  * Replaced legacy Agent Father references with Platform Operator and Human-in-the-Loop guidance.
  * Removed obsolete tutorials and deep-dive documentation.

* **Chores**
  * Removed automatic initial-agent import and related setup instructions.
  * Updated demo guidance to reflect manual agent creation.


ginccc:
Pushed 5da87d7 addressing the CI failure and both of Copilot's suppressed comments:

- **Build & Test failure**: `RestImportServiceHelpersTest` and `RestImportServiceUncoveredBranchTest` construct the service **via reflection** (`newInstance(null × 9)`), which the constructor narrowing missed — compiles fine, `wrong number of arguments: 9 expected: 7` at runtime. Both helpers now size the argument array from `constructor.getParameterCount()`, so the next signature change cannot re-break them. 91 tests across the two classes pass locally.
- **Copilot suppressed comment 1** (agent descriptor): the rebrand changed `name` but the `description` still said "EDDI's built-in agent creation wizard" — the exact shipped-product claim this PR removes. Rewritten to describe it as a non-shipping reference fixture.
- **Copilot suppressed comment 2** (property descriptor): "auto-vault for API keys" was wrong — the property setter deliberately uses `scope: "conversation"` and delegates vaulting to the receiving setup API (AGENTS.md §5.6). Rewritten. The wizard's in-conversation line ("auto-encrypted in the vault when available") stays: it accurately describes the receiving setup API's behavior.

ginccc:
Thanks — worked through all 13. Pushed 22fbb20: **7 fixed, 6 declined.**

### Fixed (all genuinely introduced or worsened here)

| Finding | Outcome |
|---|---|
| `README.md:41` stale "starter agent" | **Fixed.** Line 413 also promised `EDDI + database + starter agent via Docker`. Neither string contains "Agent Father" — which is exactly why my original sweep missed them. A removal sweep has to grep for what the thing *did*, not only what it was called. |
| `README.md:469` duplicate HITL row | **Fixed.** Line 456 already had one, with a better description. Row dropped rather than reworded. |
| Provider count 12 vs 11 | **Fixed**, and explained rather than just corrected: the chooser splits `gemini`/`gemini_vertex` while the "12 providers" figure folds in OpenAI-compatible endpoints (DeepSeek, Cohere). AGENTS.md §5.6 now says why the two numbers differ and that the fixture is not a provider catalogue. |
| `AGENTS.md:627` narrow the test guarantee | **Fixed** — and you were right in a way I could measure: the sweep's own output reads `32 configs checked, 24 skipped (patch/descriptor/unmapped)`. "Keep it parseable and save-time-valid" claimed far more than the tests deliver. Reworded to state what a green sweep actually means. |
| `docs/httpcalls.md:25` apicalls vs HttpCalls | **Fixed** by stating the duality explicitly (`apicalls` in store/URI, `httpcalls` in workflow step/extension, per AGENTS.md §5.5) rather than picking one name and leaving `architecture.md` looking like it describes something else. |
| Reflection helpers select `[0]` | **Fixed.** Deriving the *arity* dynamically fixed the crash but not the *selection*. Both now assert exactly one declared constructor first, so an added overload fails with a clear message. |

Also caught while acting on the provider-count comment: my first draft of that AGENTS.md note asserted the fixture "predates one provider." I hadn't verified it and it was wrong — removed before commit.

94 tests pass across the four affected classes; `validate` clean.

### Declined — all pre-existing in a fixture this PR only *renamed*

Five findings are in `docs/agent-configs/rule-based-reference/`, whose content is unchanged by this PR apart from descriptor text; `git log --follow` shows it as a pure rename:

- missing free-text fallback in the provider-selection rules
- plaintext `apiKey` in conversation scope — this is the **deliberate documented §5.4 pattern** (delegate vaulting to the receiving service), and §5.6 already calls it out; switching to `scope: "secret"` here would contradict the docs
- retry configured on a non-idempotent `POST /setup`
- no `${caller:token}` header on a call to EDDI's own API (§5.4 does recommend it)
- unescaped property interpolation in a hand-assembled JSON body

Plus `install.ps1:757` (no UTF-8 BOM with 16 emoji): verified against `main` — `3c 23` there too, no BOM. Pre-existing, and this PR *removed* two of those emoji rather than adding any.

I've queued these as a **separate task** rather than dropping them. The rename arguably raises the stakes, since AGENTS.md §5.6 now points at this config as explicitly canonical — but fixing agent behaviour is a different change from removing a feature, and mixing them would make both harder to review.

rolandpickl (approved):

PR #675, opened 2026-08-11
fix: audit ledger self-tamper defect, plus review fixes across CI, docs and code hygiene

ginccc:
A full critical review of the repository, and the fixes for what it found.

Most of what the review looked for was not there — no swallowed exceptions, no non-thread-safe
statics, no mutable state in the singleton lifecycle tasks, zero `@Disabled` tests across 14,000+.
The findings are few, but one of them matters, and the pattern in the rest is worth naming: nearly
every problem sat in something no automated check covered.

## The one that matters

**The audit ledger could report itself as tampered under load.**

`AuditLedgerService` caps its sequence table at 50,000 conversations and used to `clear()` the whole
thing on overflow, on the stated reasoning that re-seeding from `countByConversation` was *"correct,
only slower"*. It is not. Entries sit in the in-memory queue for up to a flush interval — longer
while a failing store is retried — so a conversation with queued entries has consumed chain
positions the store cannot see yet. Re-seeding from the store count **hands the same position out
twice**, and the verifier grades a duplicate exactly like a gap:

> *"Duplicates matter as much as gaps… Reporting INTACT here would hand an auditor a false
> assurance."* — `RestAuditStore.checkChain`

The `undelivered` table exists so the ledger's own back-pressure cannot read as tampering, but it
only exculpates **gaps**; duplicates had no such channel. On a busy deployment the queue is never
empty, so essentially every overflow produced them → `ChainStatus.BROKEN`, i.e. the ledger accusing
the deployment of deleting audit records because its own bookkeeping wrapped around.

Now only counters whose positions are all accounted for — persisted, or attributed in `undelivered`
— are evicted. A `ReentrantReadWriteLock` spans "position consumed" → "entry visible in the queue"
against eviction (that window contains HMAC and Ed25519 signing). `flush()` publishes its in-flight
batch, because between the poll and a successful append those positions are in neither queue nor
store. Still-full-after-eviction yields `UNSEQUENCED`, degrading the window to `UNAVAILABLE`
("cannot be established") rather than to an accusation.

## Also fixed

- **`HUMAN_DECIDES` blamed a feature that ships.** The rejection read *"needs human group members
  (I6), which are not available yet"* — ~150 lines below the same file's I6 matrix that validates
  those members. They shipped in 10c; the real blocker is the missing resume path for a paused
  tie-break.
- **`SafeHttpClient` documented a guarantee it did not provide.** It claimed an overall wall-clock
  timeout across all hops, but the budget is only checked *between* hops, so a hop that trickles its
  body hung forever. Both initial and redirect hops are now bounded.
- **A rate-limit bucket could latch shut permanently** — `elapsedNanos * limit` overflowed in long
  arithmetic after ~107 idle days, driving tokens negative so `tryAcquire` refused every later call.
- `ConversationStepRunner` registered the in-flight conversation outside the `try` that unregisters it.

## The gaps outside the build

- **`install.sh`/`install.ps1` had no verification of any kind** — 95 KB of shipped script, the
  README's headline `curl | bash` path, in **no** path filter (so an installer-only PR skipped the
  whole pipeline) with no lint, test, or syntax parse anywhere. New `shell-lint` job.
- **Auto-approve treated an absent check as a passing one.** A merge-conflicted PR never triggers
  CI/CD, so nothing failed and it would have approved with a body claiming all checks passed on a
  commit that was never built. Required jobs are now checked by name.
- **38 broken links → zero.** 32 shared one cause: every `planning/*.md` computed `../` and `../../`
  as if it lived under `docs/`, so `../../AGENTS.md` pointed outside the repository. The rest broke
  the *onboarding* tutorials, which linked into a `.gitbook/assets/` directory that does not exist.
- **Dead code**: `CannotExecuteException` (no references) and `ILogoutEndpoint` (a JAX-RS interface
  with no implementation, advertising `/user/isAuthenticated` to OpenAPI and serving nothing).
- **A 1.4 MB `grafana.db` was committed**, in a `grafana-data/` directory orphaned by the move to a
  named Docker volume. Its `eddi-operations.json` turned out to be a *richer* dashboard than the
  provisioned one (21 panels, "Operations Command Center") being maintained while visible to nobody
  — so it was rescued into `docs/monitoring/` and wired up, rather than deleted with the rest.
- **575 inline FQNs** against AGENTS.md §4.7 (not the permitted disambiguation case — each simple
  name resolved to exactly one class, and `IConversationService` imported `java.util.List` on one
  line while writing it fully-qualified on another).

## Verification

Every fix has a regression test that was **mutation-checked** — the fix reverted, the test confirmed
to fail with the message it would show whoever broke it. Writing them found a further defect in the
eviction fix itself: `flush()` published its batch *after* draining the queue, leaving a window where
entries were in neither collection.

Two tests guard a class of bug rather than one defect: `DocumentationLinksTest` resolves every
relative link in every markdown file (link rot is invisible to every other check here — it caught one
the manual sweep missed), and `ImportStyleTest` enforces §4.7 (writing it revealed the original audit
had under-counted 141 vs the real 575, because its pattern required a segment after `java.util`).

The 575-name refactor was verified with a **clean** `test-compile` — an incremental build reuses
stale classes and hides exactly this kind of break — and by extracting every string literal from all
273 mechanically-changed files and confirming they are byte-identical to `origin/main`, so nothing
reached inside a reflective name, config key, or log format.

Local suite: 14,473 tests, and the only failures are this sandbox's inability to open loopback
sockets — identical in count and cause to the pre-change baseline. CI is the arbiter for those.


ginccc:
Closing in favour of two smaller PRs, so automated review can actually run.

At 302 files this exceeded both CodeRabbit's limit (>100) and Copilot's (>300), so neither reviewed it — and the part crowding them out was a 263-file mechanical refactor that does not need line-by-line reading.

- **#677** — the substantive fixes (41 files): the audit-ledger self-tamper defect, SafeHttpClient, ToolRateLimiter, the HUMAN_DECIDES message, CI coverage for the installers, the auto-approve absent-check gap, 38 broken links, dead code, and all the regression tests.
- **#676** — the 575 inline FQN replacements, stacked on #677.

Same commits, same green CI, reviewable.

PR #676, opened 2026-08-12
chore: replace 575 inline fully-qualified names with imports

ginccc:
**Stacked on #677** (`fix/review-defects`) — review that one first; this PR's diff is against it.

Split out of #675 so the substantive fixes there could be reviewed at all: at 302 files both
CodeRabbit (>100) and Copilot (>300) declined. This is the half that does not need line-by-line
reading.

## What

AGENTS.md §4.7 asks for a top-level import and the simple name. These were **not** the permitted
disambiguation case — `PendingApprovalSummary`, `HitlDecision`, `ToolApprovalsConfig`,
`ConversationMemorySnapshot` and `ControlSignal` each resolve to exactly one class, and
`IConversationService` imported `java.util.List` on line 17 while writing `java.util.List<…>`
fully-qualified on line 355.

The two genuine cases — `mongo.HistorizedResourceStore extends datastore.HistorizedResourceStore`
and its `Modifiable` twin — are detected and left alone, and named explicitly in the test's
allowlist so adding to it is a reviewable act rather than silent drift.

## Why the test is in this PR

`ImportStyleTest` is what stops this re-accumulating. They build up precisely because nothing fails
when one is added: the code compiles either way, so the rule was advice a reviewer had to catch by
eye. Writing it also showed the original audit had **under-counted** — its pattern required a
package segment after `java.util`, so `java.util.List` never matched. The real total was 575, not
the 141 first reported.

## Verification

- **Clean** `test-compile`, not incremental — a reused stale class hides exactly this kind of break
  in a caller that was never edited.
- **Every string literal in all changed files compared against `origin/main`: byte-identical.** This
  is the check that matters for a rewrite this broad. A substitution that reached inside a literal —
  a reflective class name, a config key, a log format — would still compile, still pass every test,
  and show up nowhere else.
- Full suite green in CI.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -7,6 +7,41 @@
 
 
 
+## 🔤 chore: replace 575 inline fully-qualified names with imports (2026-08-12)
+
+**Repo:** EDDI (`chore/inline-fqn-cleanup`, stacked on `fix/review-defects`)
+
+The mechanical half of the repository review, split from it so the substantive fixes could be
+reviewed at all — at 302 files both CodeRabbit (>100) and Copilot (>300) decline outright, and this
+is the part that does not need line-by-line reading.
+
+AGENTS.md §4.7 asks for a top-level import and the simple name. These were **not** the permitted
+disambiguation case: `PendingApprovalSummary`, `HitlDecision`, `ToolApprovalsConfig`,
+`ConversationMemorySnapshot` and `ControlSignal` each resolve to exactly one class, and
+`IConversationService` imported `java.util.List` on line 17 while writing `java.util.List<…>`
+fully-qualified on line 355. The two genuine cases — `mongo.HistorizedResourceStore extends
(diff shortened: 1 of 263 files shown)


ginccc:
**Note on CI**: `ci.yml` triggers on `pull_request: branches: [main]`, so the build does not run against this PR while it targets `fix/review-defects`. It will run automatically once #677 merges and this retargets to `main`.

In the meantime the same file set was verified:
- The identical changes ran the **full pipeline green** in #675 (Build & Test, Integration Tests, CodeQL, Trivy, Gitleaks, Fuzz) before that PR was split.
- Locally: **clean** `test-compile` (not incremental — a reused stale class hides exactly this kind of break in an unedited caller) and the full suite at baseline.
- Every string literal in all changed files compared against `origin/main`: byte-identical.
- Copilot reviewed all 262 files and generated no comments.

Merge #677 first; this one should then be re-checked by CI on its own.

aisabella-ai (approved):

PR #677, opened 2026-08-12
fix: audit ledger self-tamper defect, plus review fixes across CI, docs and hygiene

ginccc:
A full critical review of the repository, and the fixes for what it found. Split out of #675 so
automated review can actually run — that PR carried a 263-file mechanical refactor alongside these
changes, and both CodeRabbit (>100 files) and Copilot (>300) declined to review it. The refactor is
stacked in #676.

Most of what the review looked for was not there — no swallowed exceptions, no non-thread-safe
statics, no mutable state in the singleton lifecycle tasks, zero `@Disabled` tests across 14,000+.
The findings are few, but one matters, and the pattern in the rest is worth naming: nearly every
problem sat in something no automated check covered.

## The one that matters

**The audit ledger could report itself as tampered under load.**

`AuditLedgerService` caps its sequence table at 50,000 conversations and used to `clear()` the whole
thing on overflow, on the stated reasoning that re-seeding from `countByConversation` was *"correct,
only slower"*. It is not. Entries sit in the in-memory queue for up to a flush interval — longer
while a failing store is retried — so a conversation with queued entries has consumed chain
positions the store cannot see yet. Re-seeding from the store count **hands the same position out
twice**, and the verifier grades a duplicate exactly like a gap:

> *"Duplicates matter as much as gaps… Reporting INTACT here would hand an auditor a false
> assurance."* — `RestAuditStore.checkChain`

The `undelivered` table exists so the ledger's own back-pressure cannot read as tampering, but it
only exculpates **gaps**; duplicates had no such channel. On a busy deployment the queue is never
empty, so essentially every overflow produced them → `ChainStatus.BROKEN`: the ledger accusing the
deployment of deleting audit records because its own bookkeeping wrapped around.

Eviction now drops only counters whose positions are all accounted for — persisted, or attributed in
`undelivered`. A `ReentrantReadWriteLock` spans "position consumed" → "entry visible in the queue"
against eviction (that window contains HMAC and Ed25519 signing). Still-full-after-eviction yields
`UNSEQUENCED`, degrading the window to `UNAVAILABLE` ("cannot be established") rather than to an
accusation.

**Writing the tests found a second defect in that fix**: `flush()` published its in-flight batch
*after* draining the queue, leaving a window where entries were in neither collection. Drain and
publish now happen under the same lock.

## Also fixed

- **`HUMAN_DECIDES` blamed a feature that ships.** The rejection read *"needs human group members
  (I6), which are not available yet"* — ~150 lines below the same file's I6 matrix that validates
  those members. They shipped in 10c; the real blocker is the missing resume path for a paused
  tie-break.
- **`SafeHttpClient` documented a guarantee it did not provide.** It claimed an overall wall-clock
  timeout across all hops, but the budget is only checked *between* hops, so a hop that trickles its
  body hung forever. Both initial and redirect hops are now bounded.
- **A rate-limit bucket could latch shut permanently** — `elapsedNanos * limit` overflowed in long
  arithmetic after ~107 idle days, driving tokens negative so `tryAcquire` refused every later call.
- `ConversationStepRunner` registered the in-flight conversation outside the `try` that unregisters it.

## The gaps outside the build

- **`install.sh`/`install.ps1` had no verification of any kind** — 95 KB of shipped script, the
  README's headline `curl | bash` path, in **no** path filter (so an installer-only PR skipped the
  whole pipeline) with no lint, test, or syntax parse anywhere. New `shell-lint` job; `install.sh`
  turned out already clean at ShellCheck's warning severity, so the gate lands green.
- **Auto-approve treated an absent check as a passing one.** A merge-conflicted PR never triggers
  CI/CD, so nothing failed and it would have approved with a body claiming all checks passed on a
  commit that was never built. Required jobs are now checked by name. Docs-only PRs still approve —
  a job skipped by its `if` does report a check-run.
- **38 broken links → zero.** 32 shared one cause: every `planning/*.md` computed `../` and `../../`
  as if it lived under `docs/`, so `../../AGENTS.md` pointed outside the repository. The rest broke
  the *onboarding* tutorials, which linked into a `.gitbook/assets/` directory that does not exist;
  those now import EDDI's own `/openapi` into Postman, which cannot rot the way a committed
  collection did.
- **Dead code**: `CannotExecuteException` (no references anywhere) and `ILogoutEndpoint` (a JAX-RS
  interface with no implementation, advertising `/user/isAuthenticated` to OpenAPI and serving
  nothing).
- **A 1.4 MB `grafana.db` was committed**, in a `grafana-data/` directory orphaned by the move to a
  named Docker volume. Its `eddi-operations.json` turned out to be a *richer* dashboard than the
  provisioned one (21 panels, "Operations Command Center") being maintained while visible to nobody
  — so it was rescued into `docs/monitoring/` and wired into the compose mount and both installers,
  rather than deleted with the rest.

## Verification

Every fix has a regression test that was **mutation-checked** — the fix reverted, the test confirmed
to fail with the message it would show whoever broke it:

| Fix | Reverting it produces |
| --- | --- |
| Sequence eviction | `expected <[0, 1]> but was <[0, 0]>` — the duplicate itself |
| In-flight batch retention | `expected <1> but was <-1>` (UNSEQUENCED) |
| Rate-limiter overflow | a bucket that denies every call after a long idle |
| `HUMAN_DECIDES` message | fails if it claims HUMAN members are unavailable |
| MCP class-list drift | names the `Mcp*Tools` class missing from `TOOL_CLASSES` |

`DocumentationLinksTest` guards a class of bug rather than one defect — link rot is invisible to
every other check here, which is how 38 accumulated. It earned its place immediately: CI caught a
link the local run could not, because `../security.md` resolves to the root `SECURITY.md` on
case-insensitive Windows. The link was wrong (its own text said `docs/security.md`); the test now
compares on-disk casing too, so every platform gives CI's verdict.

## Summary by CodeRabbit

* **New Features**
  * Added an EDDI operations dashboard for Grafana, covering traffic, conversations, tools, costs, quotas, messaging, and infrastructure metrics.
  * Monitoring setup now installs the operations dashboard automatically.

* **Bug Fixes**
  * Improved audit reliability during concurrent processing and recovery.
  * Added safe default HTTP response timeouts.
  * Fixed rate-limit handling after extended idle periods.
  * Clarified vote-policy validation messages.
  * Improved monitoring installation failure handling.

* **Documentation**
  * Updated API guidance to use OpenAPI and Swagger UI.
  * Corrected documentation links and expanded release notes.


ginccc:
Addressed in 6714f89 — good catch, and it was measurable.

`markdownFiles()` used `Files.walk(root)` with the `SKIPPED_DIRS` filter applied afterwards, so it paid the full traversal cost of everything it then discarded. That matters more here than the general case: `target/` always exists when the tests run and holds tens of thousands of class files, `.git/` is comparably large, and this repository's own agent worktrees live under `.claude/` — so the walk was also recursing through nested checkouts of the repository itself.

Now `Files.walkFileTree` with `SKIP_SUBTREE`. The test drops from ~1.9–4.4s to **~0.6s**, and detection is unchanged — I re-ran the mutation check (revert a link fix; it still fails naming the file and target).

ginccc:
@coderabbitai review

ginccc:
All seven addressed in 313f382. Two were real defects, not polish — thank you.

**Eviction rescans could stall the submit path** (the Major one). You were right that a full table of live conversations makes every subsequent submit for an unseen conversation take the write lock and traverse the whole queue to reach the same conclusion, with all other submitters blocked behind it. Only a flush can change the answer, so there is now a barrier keyed on a flush-generation counter. Mutation-checked: removing it turns **1 futile scan into 55** in the new `futileEvictionScanIsNotRepeatedUntilFlush`.

On `undelivered` never shrinking — correct, and deliberate: it is an *exculpatory* record, so dropping an entry can only make a later verdict stricter. It is bounded at 10k, and the barrier means the resulting "nothing evictable" state now costs one scan per flush rather than one per submit.

**Monitoring downloads failing soft** — agreed, and worse than it looks: each of those files is bind-mounted as a *file*, so a missing one makes Docker create a **directory** at the mount path and Grafana then fails to provision, including the dashboards that did download. Now fatal in both installers, which matches the reasoning already written a few lines below for the Keycloak realm.

**Both `DocumentationLinksTest` holes were real:**
- The repo-root confinement gap is the exact mistake that put 32 `planning/` links outside the repository in the first place — `normalize()` collapses `..` but does not confine. Now rejected explicitly.
- Making the SUMMARY check recursive immediately found `docs/templates/baa-template.md` unreachable from the table of contents. Now listed.

**The changelog claim** — fair. Several fixes are not mutation-checked (SafeHttpClient's absent-bound, the `Error`-only `ConversationStepRunner` window, behaviour-preserving `ThreadLocalRandom`, and the CI/dashboard changes verified by the pipeline itself). The sentence is narrowed and each exception is now stated with its reason rather than left to imply more than was done.

Also fixed: `scripts/preflight-local.ps1` was as unlinted as the installers, so the PowerShell steps now cover `scripts/` too; and the OpenAPI import instructions no longer hard-code `localhost`.

Full suite re-run: unchanged from baseline (the only failures are this sandbox's loopback sockets, which CI covers).

ginccc:
@coderabbitai review

ginccc:
Picked up the four **suppressed comments** from the latest Copilot review — thanks, two were real bugs. Addressed in 3267b75.

**`install.sh` `rm -f` (real bug, mine).** You're right, and I reproduced it before fixing: what's in the way is most likely a *directory* — the stale mount path a previous failed run left behind — and `rm -f` can't remove one. Under `set -euo pipefail` the failed `rm` becomes the script's exit point, so the user gets `rm: cannot remove ...: Is a directory` instead of the message telling them what to do, and the stale path survives to break the next run too. `rm -rf` restores it. `install.ps1` had the same shape — `Remove-Item -Force` doesn't remove a directory without `-Recurse`.

**Changelog placement (real, and a rule violation).** Correct — AGENTS.md rule 8 requires the entry on the branch carrying the work. Splitting #675 left the 575-name/`ImportStyleTest` claims here, where they don't ship. Removed from this branch and added to #676's own entry (0a504a3).

**The non-atomic capacity check.** Confirmed, and deliberate — now said out loud rather than implied. Concurrent submitters share the read lock, so N can each observe one slot left and each insert. The overshoot is bounded by concurrency, self-corrects at the next eviction, and costs one `AtomicLong` per excess conversation. Making it exact means serialising the insert path on every new conversation to enforce a bound that's a memory heuristic, not a correctness property. Documented at both the check and the declaration; the log line no longer says "full".

**The undelivered pin — retention is real, the consequence isn't, and the suggested fix is unsound.** Worth spelling out:

- *Unsound:* re-seeding a conversation with a dead-lettered gap can't work. `countByConversation` counts **persisted rows**, so with sequences 0–9 where 3 and 5 never landed, the count is 8 while the next free position is 10 — a re-seed hands out 8 and 9 again. Accounting for the highest known undelivered position doesn't rescue it either: `max(8, 6)` is still 8. A sound re-seed needs a `maxSequence(conversationId)` that `IAuditStore` doesn't expose.
- *Not reachable:* `undeliveredTracked` counts **sequences**, not conversations, and caps at `MAX_TRACKED_UNDELIVERED` = 10,000. Worst case (one sequence each) that pins 10,000 conversations out of a 50,000 table — 80% stays evictable, so "every new conversation remains UNSEQUENCED until restart" can't happen.

So the pin stays, but the reasoning is now in the code instead of my head, and `undeliveredPinCannotExhaustTheTable` asserts the headroom — raising either cap past the other fails the build rather than silently stranding conversations. That's the edit your comment actually protects against.

Suite re-run at baseline; `install.sh` still ShellCheck-clean.

ginccc:
Addressed in 580d442 — the flag was correct about the shape, and the suggested replacement is genuinely stronger.

Both operands were compile-time constants, so `10_000 < 50_000` folds to `true` and the assertion executed nothing. It *would* still have failed if someone edited a constant (the test recompiles against the new value), but that is a weak thing to rest an invariant on — and, more to the point, **it could not be mutation-checked at all**, which is how every other guard on this branch was validated.

It is now the behavioural test you described: a store outage dead-letters a conversation so position 0 is consumed but never persisted, the store recovers, the table fills and flushes. It then asserts the persisted conversations are reclaimed while the pinned one is retained, that a later conversation still receives a real chain position rather than `UNSEQUENCED`, and that the pinned chain resumes at **1** rather than reusing the dead-lettered **0**.

That last assertion is the one arithmetic could never make. Mutation-checked by dropping `undelivered.keySet()` from the retain set:

```
the counter survived, so the chain resumes past the dead-lettered 0 rather than
reusing it ==> expected: <1> but was: <0>
```

which is exactly the duplicate this mechanism exists to prevent.

ginccc:
For the record: the `Useless comparison test` comment above is pinned to commit `3267b75`, which is one commit behind the fix. The bot analysed that revision while `580d442` was landing.

Verified against current `HEAD` (`580d442`): no constant-vs-constant assertion remains. The only surviving references to `MAX_TRACKED_CONVERSATIONS` in the test file are loop bounds (`i < MAX + 5`), and every assertion in `undeliveredPinCannotExhaustTheTable` is now behavioural:

```
assertEquals(Set.of(0L), service.undeliveredSequences("pinned"), ...)   // precondition
assertNotEquals(AuditEntry.UNSEQUENCED, fresh.sequence(), ...)          // not stranded
assertEquals(0L, fresh.sequence())                                      // chains normally
assertEquals(1L, pinnedSecond.sequence(), ...)                          // 0 not reused
```

No action needed on this one.

aisabella-ai (approved):

PR #702, opened 2026-08-19
fix(secrets): redact only the value, so a redacted JSON body is still JSON

ginccc:
`SecretRedactionFilter`'s generic rule runs over the raw request **text**, and it matched the key's closing quote, the colon **and** the value's opening quote — then replaced all three along with the value (`"$1=" + REDACTED`). An ordinary body came back malformed:

```
{"modelName":"x","apiKey":"sk-ant-…"}   →   {"modelName":"x","apiKey=<REDACTED>"}
{"token":12345678,"n":1}                →   {"token=<REDACTED>,"n":1}
```

A bare string where a key/value pair was. The `sk-…` and `Bearer …` rules replace only the value and leave the document valid; this one did not — and it runs **last**, so it re-mangled what those had already redacted correctly.

## Why it matters

Every reader of a redacted body parses it, and both of the Manager's failed silently. Surfaced from [EDDI-Manager#173](https://github.com/labsai/EDDI-Manager/pull/173):

- the approval diff for a gated whole-document `PUT` fell back to comparing raw text and rendered **every line of the stored config as deleted** against the proposed body as one added line;
- worse, `detectEscalationFlags` runs *every* capability-grant check behind a `JSON.parse`. So a request that embedded a credential **and** granted `dynamicAgents.allowCreation` warned about the credential alone — and an approver reads "no second warning" as "no capability grant", which is the exact false negative that check exists to prevent.

The Manager shipped a client-side repair for the mangled shape. This is the fix at the source; that repair stays as tolerance for older backends and goes inert against a fixed one.

## Three defects, not one

**1. The output isn't JSON** — above.

**2. A secret containing a delimiter was only redacted up to it.** The value class stops at `,`, whitespace, `;`, `{`, `}` and `]`, so the tail survived into the "redacted" output:

```
{"password":"abcdefgh,SURVIVING-TAIL"}  →  {"password=<REDACTED>,SURVIVING-TAIL"}
```

**3. Found by a review pass on this branch, after a first draft.** A throwaway probe printing the filter's real output for a dozen shapes found four more leaks in my own fix — two decisions I had reasoned about and dismissed as theoretical:

| Input | First draft | Defect |
|---|---|---|
| `"password":"abcdefgh'xyz"` | `"<REDACTED>'xyz"` | closing quote needn't match the opening one |
| `"password":"it's-a-secret"` | *unchanged* | apostrophe cut it to 2 chars, under the 8-char floor |
| `"secret":"the key sk-ant-<REDACTED> is here"` | *unchanged* | "marker anywhere" guard skipped it |
| `"password":"my<REDACTED>pass"` | *unchanged* | same guard — a **bypass** for anyone who knows the marker |

## The fix

The generic rule becomes three, each replacing the value and nothing else:

| Shape | Rule | Result |
|---|---|---|
| `{"apiKey":"sk-…"}` | quoted value, runs to its **closing quote** | `{"apiKey":"sk-ant-<REDACTED>"}` |
| `{"token":12345678}` | no quotes of its own → marker takes the key's quote style | `{"token":"<REDACTED>"}` |
| `?api_key=…`, `password: …` | not JSON → separator preserved, no quotes invented | `?api_key=<REDACTED>` |

- **The closing quote is a backreference to the opening one** (`\4`), not "any quote". That fixes rows 1–2 above, and lets the value class admit the *other* quote character — a single-quoted value containing `"` gets the mirror fix free.
- **`NOT_ALREADY_REDACTED` is anchored to the start of the value**, fixing rows 3–4. It exists so the later rules don't strip the `sk-ant-` / `Bearer ` prefix an earlier rule deliberately kept — information an approver uses, and something the old rule destroyed on every named field.
- **The `${vault:…}` carve-out is now explicit.** It survived only as a side effect of the value class excluding braces; the quoted rule runs past braces, so it carries `(?!\$\{(?:vault|eddivault):)` as a stated rule — which also keeps it from being lost the next time that class is tuned.

## Deliberately not changed

- **`&` is still not a value delimiter.** It would tidy query-string redaction but would cut short every secret *containing* `&` in every other context and leak the tail. Over-redacting a raw URL that happens to sit in a log line is the safer trade, and real request URIs never reach the filter whole — `RequestRedactor` scans each query parameter's value on its own. Pinned as a test so nobody "fixes" it.
- **An escaped quote inside a value ends it early.** `\"` is a terminator in an escaped-JSON body and a literal in a plain one, and the filter cannot tell which it is reading. Resolved in favour of the escaped body — the other reading runs past the real end of the field and eats the document. Pinned as a known limit; still an improvement, since the old rule stopped at the first space and redacted nothing there.

## Verification

- `SecretRedactionFilterTest` **14 → 40** cases. Three new nested classes: `RedactedJsonStaysJson` (parses every redacted result with Jackson), `ASecretIsRedactedInFull` (parameterised over all six delimiters, plus the quote-pairing and truncated-body cases), `AlreadyRedactedValuesKeepTheirPrefix` (including both guard-bypass regressions). Plus idempotency, the 8-char floor, and neighbouring-field-not-swallowed.
- `RequestRedactorTest`, `ResolvedRequestTest`, the three `ApiCallExecutor*` suites, `ConversationMemoryUtilitiesHitlTest`, `LifecycleManagerErrorClassificationTest`, `SlackToolPauseNotificationTest`, `RestAgentEngineToolPauseDetailsTest` — all pass unchanged.
- `./mvnw validate` (Checkstyle) and `./mvnw compile` clean.

**Local full-suite caveat:** `./mvnw test` reports failures in `WebSearchToolTest`, `SafeHttpClientTest`, `SlackWebApiClientTest` (loopback sockets — the sandbox limitation AGENTS.md documents) and `DocumentationLinksTest` (it scans an untracked local `.history/` folder). Confirmed pre-existing by stashing these changes and re-running those classes: identical failures. CI is the source of truth for them.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved secret redaction across JSON and non-JSON content.
  * Preserved valid JSON structure while fully masking secrets containing delimiters.
  * Retained credential prefixes and vault placeholders where applicable.
  * Improved handling of quoted, unquoted, escaped, and already-redacted values.
  * Prevented masking from crossing quote boundaries or altering unrelated content.

* **Documentation**
  * Added a changelog entry describing the redaction improvements, compatibility updates, known limitations, and test coverage.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -7,6 +7,87 @@
 
 
 
+## 🔒 fix(secrets): redaction preserves the JSON it redacts, and stops cutting secrets short (2026-08-19)
+
+**Repo:** EDDI (`fix/redaction-preserves-json`)
+
+`SecretRedactionFilter`'s generic rule ran over the raw request TEXT and matched the key's closing
+quote, the colon **and** the value's opening quote — then replaced all three along with the value
+(`"$1=" + REDACTED`). So an ordinary body came back malformed:
+
+```
+{"modelName":"x","apiKey":"sk-ant-…"}   →   {"modelName":"x","apiKey=<REDACTED>"}
+{"token":12345678,"n":1}                →   {"token=<REDACTED>,"n":1}
+```
+
+— a bare string where a key/value pair was. The `sk-…` and `Bearer …` rules replace only the value
+and leave the document valid; this one did not, and it runs **last**, so it re-mangled what those
+had already redacted correctly.
+
+**Every reader of a redacted body parses it, and both of the Manager's failed silently.** Reported
+from EDDI-Manager PR #173: the approval diff for a gated whole-document `PUT` fell back to comparing
+raw text and rendered every line of the stored config as deleted against the proposed body as one
+added line. Worse, `detectEscalationFlags` runs *every* capability-grant check behind a `JSON.parse`,
+so a request that embedded a credential **and** granted `dynamicAgents.allowCreation` warned about
+the credential alone — and an approver reads "no second warning" as "no capability grant", which is
+the exact false negative that check exists to prevent.
+
+**A second defect, found while fixing the first.** The value class stops at `,`, whitespace, `;`,
+`{`, `}` and `]`, so a secret *containing* one of those was redacted only up to that character and
+the tail survived into the "redacted" output: `"password":"abcdefgh,SURVIVING-TAIL"` became
+`"password=<REDACTED>,SURVIVING-TAIL"`. That is a leak, not a formatting problem.
+
+**The generic rule is now three rules**, replacing the value and nothing else:
+
+| Shape | Rule | Result |
+| --- | --- | --- |
+| `{"apiKey":"sk-…"}` — quoted value | runs to the **closing quote**, keeps both quotes | `{"apiKey":"sk-ant-<REDACTED>"}` |
+| `{"token":12345678}` — no quotes of its own | marker takes the **key's quote style** | `{"token":"<REDACTED>"}` |
+| `?api_key=…`, `password: …` — not JSON | separator put back, no quotes invented | `?api_key=<REDACTED>` |
+
+Design decisions:
+
+- **The quoted rule ends the value at its closing quote, not at the first delimiter.** That is what
+  closes the partial-redaction leak; a secret with a comma or a space in it is now redacted whole.
+- **Two negative lookaheads guard it, and they deliberately use backtracking quantifiers** where the
+  rest of the file is possessive (ReDoS). `(?![^"']*<REDACTED>)` asks "does the marker appear
+  anywhere before the closing quote"; a possessive class there consumes the marker itself, finds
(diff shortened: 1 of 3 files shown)


ginccc:
Superseded by #703 — same tree, clean history.

The Secret Scanning job flagged a JWT-shaped test fixture (`RuleID: jwt`) on a newly added line. The identical string already sits in the base file at `SecretRedactionFilterTest` line 88, but the scan walks the commits a PR introduces, so a new occurrence fails while the pre-existing one passes.

Removing it in a follow-up commit did not clear the finding — the blob stays in the scanned range — and `.gitleaksignore`/`.gitleaks.toml` are deliberately taken from the base branch, so neither can switch it off from inside the PR. Force-push is banned by AGENTS.md rule 4, so a fresh branch was the only forward-only option.

#703 replaces the fixture with an opaque bearer token (the assertion is about prefix preservation, which does not depend on the token's structure; `redact_bearerToken` still covers the JWT path) and verifies clean with gitleaks 8.30.1 locally.

PR #724, opened 2026-08-29
fix(review): close an SSRF gap, and two tests that passed for the wrong reason

ginccc:
A review pass for dead code, defects and thin test coverage. Three real problems and one
under-enforced guard; each behavioural fix is **mutation-verified** (revert the fix, confirm
the new test fails).

Dead code came up essentially empty — every candidate turned out framework-wired
(`OpenApiTagSortFilter` via Quarkus `@OpenApiFilter`, `LifecycleModule` as a CDI producer,
`URIMessageBodyProvider` as a JAX-RS `@Provider`), there are zero `TODO/FIXME` markers in
`src/main`, and no `ILifecycleTask` holds mutable instance state.

---

## 1. `SourceUrlValidator` accepted internal hosts the rest of the codebase refuses

The remote agent-sync endpoints (`backup/import/sync*`) validated `sourceUrl` with a second,
local copy of the SSRF address predicate built from the four JDK checks. Those do not cover:

| Range | Why the JDK misses it |
| --- | --- |
| RFC 4193 IPv6 ULA `fc00::/7` | `isSiteLocalAddress()` only matches the deprecated `fec0::/10` |
| RFC 6598 CGNAT `100.64.0.0/10` | not a site-local range; used by Tailscale and some k8s pod CIDRs |
| IPv4 multicast `224.0.0.0/4` | not covered by any of the four predicates |

These endpoints are `@RolesAllowed({"eddi-admin", "eddi-editor"})`, so the gap was reachable by
the **lower-privileged** role. `isPrivateIp` now delegates to
`UrlValidationUtils.isPrivateAddress` — which AGENTS.md already names as the thing to call
before fetching a user-controlled URL — instead of keeping the weaker duplicate. That is also
what §4.7 "Unification over duplication" asks for. `isPrivateAddress` is promoted to `public`
and documented as the single definition of an unsafe outbound address.

**Mutation-verified:** with the old predicate restored, `100.64.0.1`, `fd00::1`, `fc00::1` and
`224.0.0.1` were all *accepted*.

The wrapper keeps its own messages and its HTTPS-in-production rule (which has no equivalent in
`UrlValidationUtils`), so no existing message assertion changes. Deliberately **not** adopted:
`UrlValidationUtils`' `.local`/`.internal` hostname block — those hostnames resolve and are then
caught by the address check anyway, and blocking them by name would newly reject a legitimate
corporate sync target.

## 2. Two audit dead-letter tests never tested what they claimed on Linux

`AuditLedgerServiceBranchTest` forced the file-fallback **failure** branch with
`"Z:\nonexistent\path\deadletter.jsonl"`. That is unwritable only on Windows — a backslash is a
legal character in a Unix filename, so on the Linux CI runner the whole string is one relative
filename that `Files.write(..., CREATE)` happily creates.

So the two assertion-free tests exercised the *success* path on CI while their comments claimed
the failure path, and left a junk file named `Z:\nonexistent\path\deadletter.jsonl` in the build
directory — which is not gitignored.

Replaced with `@TempDir` plus a deliberately-uncreated parent directory (`CREATE` does not create
parent directories, so it throws `NoSuchFileException` on both platforms). Both tests gained real
assertions: that NATS was actually attempted (or actually skipped), and that the dead-letter file
does **not** exist afterwards — which is what makes them fail if the write starts succeeding again.

**Mutation-verified:** pointing the helper at a writable path fails exactly those two tests.

## 3. `McpToolsProvider` sat at 31% coverage, including its tool-confusion defence

`McpToolsProviderTest`'s javadoc asserted discovery was "already covered indirectly by
`AgentOrchestratorExtendedTest`". Measurement disagreed — **264 of 383 instructions and 41 of 50
branches missed**. The indirect suites drive discovery with a mocked memory whose
`getAgentVersion()` is null, so `WorkflowTraversal` returns before the per-server loop is ever
entered, and the `McpToolProviderManager*Test` suites cover the *manager*, not this class.

Left untested as a result: whitelist/blacklist filtering (the blacklist is an operator security
control), the first-write-wins collision handling the class documents at length as an
anti-tool-confusion measure, the spec-without-executor skip, the resource-bridge opt-in and its
`IllegalArgumentException` to `INVALID_CONFIGURATION` path, and the `asProviderFailures` mapping.

New `McpToolsProviderDiscoveryTest` covers all of it — **31.1% to 93.5%**, branches missed 41 to 10.
The stale javadoc is corrected, and `contribute_nullFlag_defaultsToEnabled` (which asserted only
`assertNotNull`, so passed whether or not the flag short-circuited) now verifies discovery was
actually attempted.

**Mutation-verified:** removing the collision guard fails `collisionKeepsFirstSpecAndItsExecutor`.

> Note for future test authors, recorded in the new class comment: `WorkflowTraversal` memoizes a
> completed traversal for two seconds in a **static** map keyed on
> `agentId|version|stepType|configClass`, so every test allocates its own agent id.

## 4. `ImportStyleTest` did not enforce the rule it documents

Second commit, kept separate because it is purely mechanical.

The guard for AGENTS.md §4.7 ("never inline a fully-qualified name") matched only
`ai.labs.eddi|java.util|java.time|java.nio.file`, so every third-party FQN was invisible to it —
the rule was enforced on roughly a tenth of the surface it claims. Measured blind spot:
**381 inline FQNs across 118 files** (`jakarta`, `javax`, `org.eclipse`, `org.jboss`,
`com.fasterxml`, `io.quarkus`, `io.smallrye`, `io.micrometer`, `io.nats`, `org.bson`,
`com.mongodb`, `org.postgresql`, `dev.langchain4j`, plus `java.io` / `java.net` / `java.lang` /
`java.security`).

Essentially none were the disambiguation case §4.7 permits — they were simply missing imports.
Rather than park 118 files in an allowlist (the silent drift the test's own doc warns against),
the pattern is widened to an explicit root list and the violations fixed. The root list stays
explicit rather than a general lowercase-dotted-path shape, because a generic pattern also
matches method chains and builder idioms on a lowercase receiver, which are not FQNs at all.

**One genuine collision surfaced and is allowlisted:** `NatsConversationCoordinator` imports
`io.nats.client.api.*`, which brings in `io.nats.client.api.Error`, so its
`catch (RuntimeException | java.lang.Error e)` clauses need the inline FQN — rewriting them makes
the reference ambiguous and the file stops compiling, while an explicit `import java.lang.Error`
is a redundant import Checkstyle flags. Worth recording that the automated rewrite's conflict
check consulted only *single-type* imports, so a name introduced by a wildcard import was
invisible to it; the compiler caught it.

No behaviour changes in this commit — every edit swaps an inline FQN for the identical type named
by a top-level import, or moves an import line.

---

## Verification

Measured locally against a pre-change baseline on the same machine (the local suite has a known
set of environmental failures — loopback sockets, Docker, network — per the AGENTS.md sandbox
caveat):

| | Baseline | After |
| --- | --- | --- |
| Tests run | 20,295 | **20,311** (+16) |
| Failures / Errors | 8 / 193 | **8 / 193** |
| Failing classes | 11 | **same 11, identical set** |
| Instruction coverage | 89.91% | **90.01%** |
| Branch coverage | 79.24% | **79.38%** |

The failing-class *set* was diffed, not just the counts — that catches a swap where one class
newly breaks and another newly passes. `McpToolsProvider` 31.1% to 93.5%; `SourceUrlValidator`
96.3%. Checkstyle is unchanged at its pre-existing violation count.

Verified with a **clean** `test-compile` rather than an incremental one, since a type-level
refactor otherwise reuses stale `.class` files and hides breaks in unedited callers. Rebased onto
current `main` (which had picked up the caller-supplied-connections work) and re-verified: the
widened guard passes against that newly-landed code too.

## Deliberately left alone

- `RemoteApiResourceSource` builds a raw `HttpClient` rather than using `SafeHttpClient`,
  contrary to §4.4. Real, but a distinct change with its own blast radius (redirect/timeout
  semantics differ), and not urgent — the JDK default redirect policy is `NEVER`, so there is no
  redirect-based bypass.
- The remaining measured coverage gaps, chiefly `RestImportService`'s partial-failure/rollback
  paths (1,165 missed instructions, 118 branches).

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -7,6 +7,71 @@
 
 
 
+## 🧹 refactor(style): make ImportStyleTest enforce the rule it documents (2026-08-28)
+
+**Repo:** EDDI (`claude/code-review-test-coverage-59bf99`)
+
+`ImportStyleTest` guards AGENTS.md §4.7 ("never inline a fully-qualified name"), but
+its `INLINE_FQN` pattern only matched four package roots —
+`ai.labs.eddi|java.util|java.time|java.nio.file`. Every third-party FQN was invisible
+to it, so the rule was enforced on about a tenth of the surface it claims to cover.
+
+Measured blind spot: **381 inline FQNs across 118 files**, for roots the project
+actually depends on — `jakarta`, `javax`, `org.eclipse`, `org.jboss`, `com.fasterxml`,
+`io.quarkus`, `io.smallrye`, `io.micrometer`, `io.nats`, `org.bson`, `com.mongodb`,
+`org.postgresql`, `dev.langchain4j`, plus the JDK's `java.io`, `java.net`, `java.lang`
+and `java.security`. Examples: `jakarta.ws.rs.NotFoundException` in `McpHitlTools`,
+`io.micrometer.core.instrument.Counter` as a field type in `AuditLedgerService`,
+`org.eclipse.microprofile.openapi.models.tags.Tag::getName` in `OpenApiTagSortFilter`.
+
+Essentially none were the disambiguation case §4.7 permits — they were simply missing
+imports. Rather than park 118 files in an allowlist (the test's own doc argues an
+`ALLOWED` entry should be "a deliberate, reviewable act rather than silent drift", and
+an allowlist that never shrinks is exactly the drift it warns about), the pattern is
+widened to an explicit root list and the violations are fixed.
+
+The root list stays explicit rather than a general lowercase-dotted-path shape,
+because a generic pattern also matches method chains and builder idioms on a
+lowercase receiver, which are not FQNs at all.
+
+### One genuine collision found, and allowlisted
+
+`NatsConversationCoordinator` imports `io.nats.client.api.*`, which brings in
+`io.nats.client.api.Error`. Its `catch (RuntimeException | java.lang.Error e)` clauses
+mean the JDK type, and the inline FQN is load-bearing: rewriting it to `Error` makes
+the reference ambiguous and the file stops compiling. An explicit
+`import java.lang.Error` resolves it but is a redundant import (`java.lang` is
+implicit) that Checkstyle flags — so the inline FQN really is the only clean spelling.
+Added to `ALLOWED` with that reasoning recorded.
+
+Worth noting how this surfaced: the automated rewrite's conflict check only consulted
+*single-type* imports, so a name introduced by a wildcard import was invisible to it.
+The compiler caught it. Anyone repeating this exercise should expect wildcard imports
+to hide exactly this class of collision.
+
+### Verification
+
+Clean `test-compile` (not incremental — a type-level refactor reuses stale `.class`
+files otherwise), `ImportStyleTest` green against the widened pattern, and the full
+unit suite re-run against the pre-change baseline of 20,295 tests / 8 failures /
+193 errors (all environmental: loopback sockets, Docker, network). Checkstyle is
+unchanged at its pre-existing violation count — the one violation this work did
+introduce, a redundant `import java.lang.Error`, is gone with the revert above.
+
+No behaviour changes: every edit replaces an inline FQN with the identical type named
+by a top-level import, or moves an import line.
+
+Branch totals after both commits: **20,311 tests** (+16 over the baseline: 13 new
+MCP discovery tests, 3 new SSRF tests), same 8 failures / 193 errors, and an
+*identical* set of failing classes - so nothing regressed. Coverage moved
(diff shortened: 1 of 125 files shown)


ginccc:
Superseded by #725 (the three behavioural fixes, 7 files) and #726 (the import-style refactor, 119 files).

Splitting because this PR's 125 files tripped CodeRabbit's 100-file limit, so the SSRF fix and the two tests that passed for the wrong reason got no bot review at all — the mechanical refactor was crowding out review of the part that actually changes behaviour. Same commits, no content changes beyond the changelog split.

PR #783, opened 2026-09-18
fix(rag): register ai.labs.rag so RAG workflow steps can be deployed

ginccc:
## Summary

A workflow step of type `eddi://ai.labs.rag` cannot be deployed on `main` today. `WorkflowStoreClientLibrary` resolves a step by `URI.getHost()` against the `@LifecycleExtensions` map and throws `UnrecognizedExtensionException` when the key is absent — and no module ever registered `ai.labs.rag`.

So the two knowledge-base options documented in `docs/rag.md` are undeployable, and only `httpCallRag` works end to end. `RestWorkflowStepStore` builds the Manager's step chooser from the same map, so the backend never offered the step either; the Manager's MSW fixture hard-codes it, which is why its suite stayed green.

This is independent of the RAG ingestion work it was found during — `main` needs it either way.

## Changes

- **`RagTask`** — a config carrier. `execute()` is a deliberate no-op: retrieval stays in `RagContextProvider` inside the LLM task (the Phase 8c decision stands). `configure()` resolves the referenced `RagConfiguration`, so a broken binding fails at deploy time instead of silently returning no context on the first conversation.
- **`RagModule`** — registers it under `ai.labs.rag`.
- `getType()` returns the stage name `"rag"`, not the `eddi://` URI, per the `ILifecycleTask` contract.

## How to test

- `RagWorkflowDeploymentTest` deploys a real workflow containing a RAG step through `WorkflowStoreClientLibrary`, and pins the negative case: without the registration the whole workflow is rejected.
- `RagWorkflowExtensionIT` asserts against live CDI wiring, via `RestWorkflowStepStore`, that the step is offered and that every documented step type resolves. It is an `IT` because every `@QuarkusTest` in this repo runs in the integration job.
- Mutation-checked: reverting `getType()` to the URI and the registration key to a wrong value fails four tests.

## Type of change

- [x] Bug fix

## Summary by CodeRabbit

- **New Features**
  - RAG workflow steps using `eddi://ai.labs.rag` can now be deployed and offered in workflow configuration.
  - Knowledge-base bindings are validated during deployment, with clear errors for invalid or unavailable resources.
  - RAG steps expose a knowledge-base URI configuration field and integrate with LLM-based retrieval.

- **Documentation**
  - Added documentation and changelog details for the registered RAG workflow step and its configuration.


ginccc:
Addressed the CodeRabbit nitpick on `RagModuleTest` in 49ed26324: the other extension is now registered before `configure()` runs, and the test asserts that the same instance survives (`assertSame`), so it checks that `configure()` preserves an existing registration rather than one added afterwards.

rolandpickl (approved):

PR #788, opened 2026-09-18
fix(ui): a dialog no longer takes focus from a field the user is typing in

ginccc:
## Summary

`UI Manager Checks` failed intermittently in [run 35295321603](https://github.com/labsai/EDDI/actions/runs/35295321603/job/105446603130) with two failures that look unrelated but have **one cause, in the component**:

- `share-dialog.test.tsx` › *does not let two quick Enters skip the ownership confirmation*: `Unable to find [data-testid="share-owner-warning"]`
- `create-agent-dialog.test.tsx` › *allows typing in description field*: the field was `""` after `user.type`

### Root cause

`AccessibleDialog` moved initial focus to its first focusable element (the header's Close button) inside a `requestAnimationFrame` scheduled on open. When that frame ran late, which happens on a loaded runner, it stole focus from a field the test had already clicked into. user-event sends keystrokes to `document.activeElement`, so `bob` went to the Close button. The share subject stayed empty, Enter failed validation, and the warning was never rendered, so no amount of waiting would have found it.

The same frame also overrode every `autoFocus` inside a dialog in the real UI. `CreateAgentDialog` autofocuses its Name field, and one frame later focus sat on the X.

### Changes

- **`accessible-dialog.tsx`:** the frame leaves focus alone if it is already inside the dialog, and is cancelled on cleanup. The focus trap, Escape and return-focus behaviour are unchanged.
- **`accessible-dialog.test.tsx` (new):** holds the frame and releases it by hand, so "focus reached a field first" is deterministic. Mutation-checked: removing the guard fails 2 of its 3 tests.
- **`share-dialog.test.tsx`:** the warning assertions that directly follow a user event now wait: `findByTestId` when the warning should appear, `waitFor` when it should disappear. The `waitFor` on an absence is safe because each test has just seen the warning present. **This is hygiene, not the fix.** The first attempt was only this change, and it still failed under the reproduction below.

## Verification

- **Reproduction:** stubbing `requestAnimationFrame` to a 30–150 ms timeout, the original share-dialog tests fail with the CI error and `create-agent-dialog` fails with the CI's empty value. With this fix, the 80 ms and 150 ms copies of the old tests, the new tests and `create-agent-dialog` all pass (246/246).
- **Full Manager suite with coverage:** 413 files / 6,550 tests passing locally.
- **`npm run lint`** and **`npm run typecheck`** pass.

## Not in scope

`previousFocusRef` is captured after an `autoFocus` child has already taken focus, so on close focus "returns" to that unmounted field instead of the trigger. This bug predates this PR and is separate, so it will get its own PR.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved dialog focus behavior so fields already focused by users, including automatically focused fields, remain focused when the dialog opens.
  - Dialogs still focus the Close button by default when no field is selected.
  - Improved focus handling during dialog cleanup while preserving focus trapping, Escape-key handling, and return-focus behavior.

- **Tests**
  - Added coverage for dialog focus scenarios and asynchronous share-dialog warning states.

- **Documentation**
  - Added a changelog entry describing the dialog focus improvements.


ginccc:
@coderabbitai review

rolandpickl (approved):

PR #793, opened 2026-09-19
fix(hitl): sanitize MCP get_approval_status detail=full like REST

ginccc:
## Problem

Two approver-facing surfaces return a paused conversation's full memory snapshot with `detail=full`, and they sanitized it differently:

- **REST** `GET /agents/{id}/approval-status?detail=full` → `ConversationMemoryUtilities.sanitizePendingToolCallsForApprover(snapshot)`: nulls `argumentsRaw`, `chatTranscriptJson` and `traceSoFar`, masks the request fingerprint, re-redacts `argumentsRedacted` and the request preview with the current filter.
- **MCP** `get_approval_status` with `detail=full` → `stripRequestFingerprintsForRead(snapshot)` only.

So any MCP caller the read gate admitted received the raw arguments of every pending tool call (clear-text credentials have been observed there), the full serialized LLM transcript, and the running tool trace. On deployments without OIDC, `eddi.mcp.allow-unauthenticated=true` makes that surface reachable by anything on the network that can reach EDDI.

## Fix

- `McpHitlTools.getApprovalStatus` (`detail=full`) now serializes `sanitizePendingToolCallsForApprover(snapshot)` — the same call REST makes.
- `stripRequestFingerprintsForRead` is now **private**. The divergence existed because there were two public projections to choose from; now there is one, and the compiler rejects the partial one. The Javadoc of both says every approver surface must go through the sanitizer, and that new raw-argument fields on `PendingToolCallBatch` belong there.
- `docs/hitl.md`: the approver read-scope paragraph states the sanitization applies to REST and MCP alike.

## Reproduction / tests

`McpHitlToolsTest.getApprovalStatus_detailFull_neverServesRawArgumentsOrTranscript` puts a canary in `argumentsRaw`, `chatTranscriptJson` and `traceSoFar`, calls the MCP tool with `detail=full`, serializes through the real `JsonSerialization` (a mocked serializer would hide which fields ride along), and asserts the canary is absent while `argumentsRedacted` is still served.

- Fails on `origin/main` (all three canaries in the output); passes with the fix.
- Mutation check: reverting only the production change makes it fail again.
- `compile` (Checkstyle + `formatter:validate`), the HITL/MCP/REST/memory test classes, and the repo-wide guards (`ImportStyleTest`, `DocumentationLinksTest`, `StrictBoundaryShippedConfigsTest`, `RuleSetStoreShippedRulesetsTest`, `BuildQualityGatesTest`, `ChangelogRotationTest`) pass locally.

## Group variant — no change needed

`get_group_approval_status?detail=full` (MCP) and its REST twin both return the `GroupConversation` unmodified. That document has no `PendingToolCallBatch`, no `argumentsRaw` and no LLM transcript JSON; its `transcript` is the group discussion, which is the documented content of the full view, and `hitlLastPauseFingerprint` digests task state, not a request. Per-call tool pauses inside group turns (`inGroupTurns: INBOX`) are still reserved.

## Other external readers checked (already safe)

| Reader | What it serves from the pending batch |
|---|---|
| REST `approval-status` summary (`pauseDetails`) | `argumentsRedacted`, re-redacted at read time |
| `RestConversationStore` raw log / simple log | names-only projection |
| `ConversationService.readConversation` + all say/resume responses → MCP `read_conversation`, OpenAI-compatible bridge, `ConverseWithAgentTool`, `CreateSubAgentTool` | names-only projection / tool names |
| Slack approval cards (`SlackHitlSupport`) | `argumentsRedacted`, re-redacted |
| GDPR Art. 15 export | conversation outputs only |
| HITL audit entry | `argsDigest` (SHA-256), not the arguments |
| `RestToolHistory` | step traces, owner-only; not the pending batch |
| `RestTemplatePreview` | `MemoryItemConverter` exposes no HITL fields |

Follow-up worth considering separately: the audit `argsDigest` is an unsalted SHA-256 of the raw arguments — not a leak, but the same offline-guessing reasoning that led to masking the request fingerprint applies to low-entropy arguments.

## Concurrent change

`fix/gemini-thought-signatures` (not merged yet) adds `PendingToolCallBatch.gatingAssistantMessageJson`, which embeds raw tool arguments. **Whichever PR lands second must null that field in `sanitizePendingToolCallsForApprover` and add it to this test's canary set.**

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved security for full approval-status details returned through MCP.
  * Sensitive tool-call arguments, conversation transcripts, and execution traces are no longer exposed in approval-status responses.
  * Redacted arguments and request previews are consistently re-sanitized before being displayed.
  * REST and MCP approval-status views now apply the same sanitization rules.

* **Documentation**
  * Clarified sanitization behavior for REST and MCP approval-status views.
  * Updated the September 2026 changelog archive with recent maintenance and security fixes.


ginccc:
@coderabbitai review

rolandpickl (approved):

